Issue #5, opened 2017-01-13
l10n_us_account_profile - US Accounting


max3903:
# US Accounting Profile

This module provides all the dependencies, data, parameters and configuration panel to have Odoo as an accounting software ready for US accountants and equivalent to local solutions like Quickbooks.

# 0.0.1

#7 

- [x] account_due_list

# 0.0.2

- [x] account_reversal: https://github.com/OCA/account-financial-tools/pull/445
- [x] account_payment_batch_process: https://github.com/OCA/account-payment/pull/138
- [x] account_payment_credit_card: https://github.com/OCA/account-payment/pull/139
- [x] partner_aging: https://github.com/OCA/account-payment/pull/136 by @Sudarshan1607 
- [x] partner_daytopay: https://github.com/OCA/account-invoice-reporting/pull/53 by @bodedra
- [x] account_check_printing_report_base: https://github.com/OCA/account-payment/pull/147 by @smangukiya
- [x] l10n_us_check_writing_address: https://github.com/OCA/l10n-usa/pull/8 by @Sudarshan1607 
- [x] l10n_us_form_1099: #13 by @wujenne2
- [x] l10n_us_account_profile: #11 by @Sudarshan1607 

# 0.0.3

- [x] account_banking_reconciliation: https://github.com/OCA/bank-statement-reconcile/pull/174 by @bodedra
- [ ] partner_financial_risk: https://github.com/OCA/partner-contact/pull/360
- [ ] partner_sale_risk: https://github.com/OCA/partner-contact/issues/309
- [ ] partner_stock_risk: https://github.com/OCA/partner-contact/issues/309
- [ ] l10n_us_account_profile:

JordiBForgeFlow:
Where is l10n_us_check_printing?

We implemented check printing in the DLT103 (a common standard in the US) format here:
https://github.com/OCA/account-payment/pull/121

max3903:
@jbeficent It's not public yet, but we plan to share it.

JordiBForgeFlow:
@max3903 What is the point now, if there's already the PR stated above?

max3903:
@jbeficent Let me discuss with my team. I don't have all the info yet. That's why I started this thread.

JordiBForgeFlow:
@max3903 In any case it will be good to compare both approaches and go for the best or a mix of both ;)

JordiBForgeFlow:
I think that the following are relevant:
* account_due_list
* account_due_list_days_overdue
[...]

PR #7, opened 2017-04-13
[ADD] First version of l10n_usa_account_profile

max3903:
#5


l10n-usa-account-profile/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "US Accounting",
+    "summary": "Additional features to manage your accounting in Odoo",
```
lasley: I think it would be good to include US in the summary too - the UI shows the summary in place of the name on hover IIRC, so the fact that this is US would become obscured and require a toggle back. Maybe this?

> Additional features to manage USA accounting in Odoo

lasley (approved):

PR #8, opened 2017-04-14
[ADD] l10n_us_check_writing_address

max3903:
This module allows to print your payments on pre-printed check paper. This module adds the address to the Payee section of a check.

Supported formats

* Check on top : Quicken / QuickBooks standard
* Check in the middle: Peachtree standard
* Check at the bottom: ADP standard

# Configuration

You can configure the output (layout, stubs informations, etc.) in company settings, and manage the checks numbering (if you use pre-printed checks without numbers) in the journal setting.

# Usage

To use this module, you need to:

* Go to Accounting > Purchases > Payments
* Create a new payment with a payment method (journal) configured to print check
* Check the check option, register the payment and print the check

#5
#10 
Depends on https://github.com/OCA/account-payment/pull/143


l10n_us_check_writing_address/README.rst
```diff
@@ -0,0 +1,87 @@
...
+#. Go to Accounting > Purchases > Payments
+#. Create a new payment with a payment method (journal) configured to print check
+#. Check the check option, register the payment and print the check
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/203/10.0
+
+Known issues / Roadmap
+======================
```
lasley: Remove empty section

l10n_us_check_writing_address/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import (
```
lasley: These should be on multiple lines - see [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)

l10n_us_check_writing_address/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'US Check Printing with Payee Address',
+    'summary': 'Print US Checks',
+    'version': '10.0.1.0.0',
+    'license': 'AGPL',
```
lasley: AGPL-3

l10n_us_check_writing_address/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+
+from odoo import api, fields, models, _
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def _display_address(self, without_company=False):
+        '''
```
lasley: ❤️  that you added a docstring. Please make it [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) compliant though

l10n_us_check_writing_address/models/res_partner.py
```diff
@@ -0,0 +1,43 @@
...
+        standards of the country where it belongs.
+
+        :param address: browse record of the res.partner to format
+        :returns: the address formatted in a display that fit its country habits (or the default ones
+            if not country is specified)
+        :rtype: string
+        '''
+        # get the information that will be injected into the display format
+        # get the address format
+        if self.street2 == False:
```
lasley: Hmm I think we can simplify this a bit.

Do I assume correctly that a blank line was being added if using the standard `country_id.address_format` when the `street2` was blank? If so, maybe a lil something like:

```
address_format = self.country_id.address_format
if not self.street2:
    address_format = address_format.replace('%(street2)\n', '')
```

Or maybe slightly more robust with a regex, but I think that we'd be fine seeing as this is specific to US.

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.osv import osv
```
lasley: Old API

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+from odoo import api, fields, models, _
+
+LINE_FILLER = '*'
+INV_LINES_PER_STUB = 9
+
+
+class ReportPrintCheck(report_sxw.rml_parse):
+
+    def __init__(self, cr, uid, name, context):
+        super(report_print_check, self).__init__(cr, uid, name, context)
```
lasley: s/report_print_check/ReportPrintCheck

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+from odoo.report import report_sxw
+from odoo import api, fields, models, _
+
+LINE_FILLER = '*'
+INV_LINES_PER_STUB = 9
+
+
+class ReportPrintCheck(report_sxw.rml_parse):
+
+    def __init__(self, cr, uid, name, context):
```
lasley: Old API

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo.osv import osv
+from odoo.report import report_sxw
+from odoo import api, fields, models, _
+
+LINE_FILLER = '*'
```
lasley: It would be nicer if these were contained within the `ReportPrintCheck` class IMO. Even better though would be system params or company settings.

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+        super(report_print_check, self).__init__(cr, uid, name, context)
+        self.localcontext.update({
+            'pages': self.get_pages,
+        })
+
+    def fill_line(self, amount_str):
+        return amount_str and (amount_str+' ').ljust(200, LINE_FILLER) or ''
+
+    def get_pages(self, payment):
+        """ Returns the data structure used by the template : a list of dicts containing what to print on pages.
```
lasley: PEP-257

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        stub_pages = self.make_stub_pages(payment)
+        multi_stub = payment.company_id.us_check_multi_stub
+        pages = []
+        for i in range(0, stub_pages != None and len(stub_pages) or 1):
```
lasley: I'm having a bit of trouble understanding this logic. Also, comparisons for `None` should always use `is` opposed to `==`.

I think the below would be more clear, assuming it hits the intent:

```
if stub_pages is None:
    return pages

for i in range(0, len(stub_pages) or 1):
```

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            pages.append({
+                'sequence_number': payment.check_number\
```
lasley: This conditional is too complex to be inside of a dictionary assignment. Please break out into a temporary variable

My general rule of thumb is that `\` is a dirty hack, and any time it's used is a sign of a refactor need.

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+                    else False,
+                'payment_date': payment.payment_date,
+                'partner_name': payment.partner_id.name,
+                'partner_name_addr': payment.partner_id.street,
+                'partner_name_addr2': payment.partner_id.street2,
+                'partner_city': payment.partner_id.city,
+                'partner_state': payment.partner_id.state_id.code,
+                'partner_zip': payment.partner_id.zip,
+                'currency': payment.currency_id,
+                'amount': payment.amount if i == 0 else 'VOID',
```
lasley: This `i == 0` is being performed multiple times. It would be more efficient and concise to perform once above, assigning to temp vars for use in the dict assignment.

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+                'amount_in_word': self.fill_line(payment.check_amount_in_words) if i == 0 else 'VOID',
+                'memo': payment.communication,
+                'stub_cropped': not multi_stub and len(payment.invoice_ids) > INV_LINES_PER_STUB,
+                # If the payment does not reference an invoice, there is no stub line to display
+                'stub_lines': stub_pages != None and stub_pages[i],
```
lasley:
```
stub_pages is not None and stub_pages[i]
```

But really I much prefer lazy evals unless the `None` is incredibly important:
```
stub_pages and stub_pages[i]
```

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            })
+        return pages
+
+    def make_stub_pages(self, payment):
+        """ The stub is the summary of paid invoices. It may spill on several pages, in which case only the check on
+            first page is valid. This function returns a list of stub lines per page.
+        """
+        if len(payment.invoice_ids) == 0:
```
lasley: It's best to rely on the truthiness of the recordsets instead of length:
```
if not payment.invoice_ids
```

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            return None
```
lasley: `return` does the same thing with less typing 😉

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+        else:
+            stub_lines = [{'header': True, 'name': "Bills"}]
+            stub_lines += [self.make_stub_line(payment, inv) for inv in debits]
+            stub_lines += [{'header': True, 'name': "Refunds"}]
+            stub_lines += [self.make_stub_line(payment, inv) for inv in credits]
+
+        # Crop the stub lines or split them on multiple pages
+        if not multi_stub:
+            # If we need to crop the stub, leave place for an ellipsis line
+            num_stub_lines = len(stub_lines) > INV_LINES_PER_STUB and INV_LINES_PER_STUB-1 or INV_LINES_PER_STUB
```
lasley: Inline assignment too complex

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            stub_pages = [stub_lines[:num_stub_lines]]
+        else:
+            stub_pages = []
+            i = 0
+            while i < len(stub_lines):
+                # Make sure we don't start the credit section at the end of a page
+                if len(stub_lines) >= i+INV_LINES_PER_STUB and stub_lines[i+INV_LINES_PER_STUB-1].get('header'):
```
lasley: It would be more clear to perform the math into temporary variables, then perform the if statement on the vars.

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+                    num_stub_lines = INV_LINES_PER_STUB-1 or INV_LINES_PER_STUB
+                else:
+                    num_stub_lines = INV_LINES_PER_STUB
+                stub_pages.append(stub_lines[i:i+num_stub_lines])
```
lasley: `stub_pages` will be a 2D array as it is here. Just making sure that's the intent & we don't instead want `.extend`

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+
+        return stub_pages
+
+    def make_stub_line(self, payment, invoice):
+        """ Return the dict used to display an invoice/refund in the stub
+        """
+        # Find the account.partial.reconcile which are common to the invoice and the payment
+        if invoice.type in ['in_invoice', 'out_refund']:
+            invoice_sign = 1
+            invoice_payment_reconcile = invoice.move_id.line_ids.mapped('matched_debit_ids').filtered(lambda r: r.debit_move_id in payment.move_line_ids)
```
lasley: Temp vars could help out with this & the corresponding below line lengths

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            invoice_payment_reconcile = invoice.move_id.line_ids.mapped('matched_credit_ids').filtered(lambda r: r.credit_move_id in payment.move_line_ids)
+
+        if payment.currency_id != payment.journal_id.company_id.currency_id:
+            amount_paid = abs(sum(invoice_payment_reconcile.mapped('amount_currency')))
+        else:
+            amount_paid = abs(sum(invoice_payment_reconcile.mapped('amount')))
+
+        return {
+            'due_date': invoice.date_due,
+            'number': invoice.reference and invoice.number + ' - ' + invoice.reference or invoice.number,
```
lasley: String interpolation is preferred (and more efficient) than concatenation. Also, this might be better as an explicit conditional
```
'%s - %s' % (invoice.number, invoice.reference)
```

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+            'amount_total': invoice_sign * invoice.amount_total,
+            'amount_residual': invoice_sign * invoice.residual,
+            'amount_paid': invoice_sign * amount_paid,
+            'currency': invoice.currency_id,
+        }
+
+
+class PrintCheckTop(osv.AbstractModel):
```
lasley: `models.AbstractModel`

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+    _inherit = 'report.l10n_us_check_printing.print_check_top'
+    _template = 'l10n_us_check_printing.print_check_top'
+    _wrapped_report_class = report_print_check
+
+
+class PrintCheckMiddle(osv.AbstractModel):
```
lasley: `models.AbstractModel`

l10n_us_check_writing_address/report/print_check.py
```diff
@@ -0,0 +1,135 @@
...
+    _inherit = 'report.l10n_us_check_printing.print_check_middle'
+    _template = 'l10n_us_check_printing.print_check_middle'
+    _wrapped_report_class = report_print_check
+
+
+class PrintCheckBottom(osv.AbstractModel):
```
lasley: `models.AbstractModel`

l10n_us_check_writing_address/report/print_check.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
lasley: No need for the data tag unless you're mixing noupdate 1 & 0.

lasley (changes requested):

JordiBForgeFlow:
Have you considered https://github.com/OCA/account-payment/tree/9.0/account_check_printing_report_dlt103?

Prints checks for.the DLT103 vhecl format, which is a standard in the US. And is pixel perfect.

max3903:
@jbeficent Can you post a sample check? Here is one with our module:
[Print Check (Top).pdf](https://github.com/OCA/l10n-usa/files/927176/Print.Check.Top.pdf)

JordiBForgeFlow:
@max3903 find attached a composition that show how your Print Check (Top).pdf prints over a DLT103 check. As you see, it is all misaligned.
image

IMHO every check format that can be purchased has it's own specific alignment. A new module should be created, that conforms to this alignment. We proposed https://github.com/OCA/account-payment/tree/9.0/account_check_printing_report_base as the base module that all format-specific modules should build on.

max3903:
@jbeficent Our check format does not respect a standard, I get that. But it's being used by many of our customers and answers a requirement from the market...

What do you think of:
* accepting this module as is in v10 
* @ursais doing those improvements (#10) in v10 or when migrating to v11

max3903:
@bodedra When you get a chance, can you fix the last comments from @lasley review please? Thanks!

max3903:
@bodedra 

I got this error message when creating a customer:
```
  File "l10n_us_check_writing_address/models/res_partner.py", line 31, in _display_address
    address_format = address_format.replace('%(street2)\n', '')
AttributeError: 'bool' object has no attribute 'replace'
```

max3903:
@smangukiya The module doesn't do anything. It should still provide the same check format as before but using the new dependency.

l10n_us_check_writing_address/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Ursa Information Systems <http://www.ursainfosystems.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import models
+# from . import report
```
vprime: Please remove the old import report line

vprime (commented):
Just remove the line commenting out stale code, and I think this is good.

l10n_us_check_writing_address/README.rst
```diff
@@ -0,0 +1,83 @@
...
+===================================
+US Check Writing with Payee Address
+===================================
+
+This module allows to print your payments on pre-printed check paper. This module adds the address to the Payee section of a check.
+
+Supported formats
+-----------------
+
+* Check on top : Quicken / QuickBooks standard
```
lasley: Please remove space before `:`

l10n_us_check_writing_address/README.rst
```diff
@@ -0,0 +1,83 @@
...
+* Check in the middle: Peachtree standard
+* Check at the bottom: ADP standard
+
+Installation
```
lasley: Remove empty section

l10n_us_check_writing_address/models/res_partner.py
```diff
@@ -0,0 +1,48 @@
...
+        Returns:
+            string: the address formatted in a display that fit its country
+            habits (or the default one if no country is specified)
+
+        '''
+        # get the information that will be injected into the display format
+        # get the address format
+        address_format = self.country_id.address_format
+        if not address_format:
+            address_format = "%(street)s\n%(street2)s\n%(city)s" \
```
lasley: This format string should be a class or file constant IMO

l10n_us_check_writing_address/models/res_partner.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+from odoo import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def _display_address(self, without_company=False):
```
lasley, line 19: I'm not sure how this is being used- but why not just use the built-in address widget?

lasley (dismissed):
Thanks @max3903 - comments inline.

Also, please add tests

lasley:
Wait am I missing something here? Where are the supported formats and the workflows in the ReadMe being introduced code-wise?

max3903:
@smangukiya You removed the report files here:
https://github.com/OCA/l10n-usa/pull/8/commits/7a5cd8dc3b8d2813cbe3606654b40eae2bf3be6e

Any reason?

l10n_us_check_writing_address/models/res_partner.py
max3903: @lasley To avoid empty lines if there is no data, I think.

smangukiya:
@max3903 : I remember that I have removed them because they were dependant on module called "l10n_us_check_printing" we have replace that dependency with oca module 'account_check_printing_report_base' which doesn't have those templates.

max3903:
@smangukiya We still need to provide the 3 reports (top, middle, bottom) with the correctly formatted address.

smangukiya:
@max3903 : I will note this point. Thanks!

max3903:
Ping @smangukiya @Sudarshan1607

bodedra (approved):
LGTM :+1:

smangukiya (approved):

max3903:
:+1:

mgosai (commented):
Looks good. (y)

PR #18, opened 2017-11-10
[ADD] l10n_us_product: Product weight and volume in US units

lasley:
* New module that allows you to record product weight in ounces and volume in inches cubed

### DIFF
--- a/l10n_us_product/models/product_template.py
+++ b/l10n_us_product/models/product_template.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    weight_oz = fields.Float(
+        compute='_compute_weight_oz',
+        inverse='_inverse_weight_oz',
+        help='Weight of the product, in ounces.',
+    )
+    volume_in = fields.Float(
+        compute='_compute_volume_in',
+        inverse='_inverse_volume_in',
+        help='Volume of the product, in inches cubed.',
+    )
+
+    @api.multi
+    @api.depends('weight')
+    def _compute_weight_oz(self):
+        for record in self:
+            record.weight_oz = self._convert_kilogram_ounce(
+                record.weight,
+            )
+
+    @api.multi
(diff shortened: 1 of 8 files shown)


l10n_us_product/models/product_template.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @api.multi
+    def _inverse_weight_oz(self):
+        for record in self:
+            record.weight = self._convert_kilogram_ounce(
+                record.weight_oz, to_oz=False,
+            )
+
+    @api.multi
+    def _compute_volume_in(self):
```
lasley: These calculations are probably wrong m3 != m, in != m3.

Likely calculation: `convert( (cubed ** (1.0/3.0) ) ** 3`

l10n_us_product/models/product_template.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @api.multi
+    def _inverse_volume_in(self):
+        for record in self:
+            record.volume = self._convert_meter_inch(
+                record.volume_in, to_in=False,
+            )
+
+    @api.model_cr_context
+    def _convert_kilogram_ounce(self, quantity, to_oz=True):
```
lasley, line 51: Note this is overly complex to allow for usage in other models (such as the module I'm about to submit for stock.picking)

max3903:
@lasley Your module is installed on Runbot but the view is not there.

Otherwise everything looks good to me.

For a future version, what do you think of adding an option to choose the default unit of measure for weight? I have customers who enters weights in pounds (lb), not ounces.

lasley:
I deleted all the coveralls spam & will take a look at why the hell it's still commenting in this repo (so annoying IMO).

> Your module is installed on Runbot but the view is not there.

Yeah for some reason I put `view` instead of `data` in the manifest. Brain fart, fixed in e2118d9

> For a future version, what do you think of adding an option to choose the default unit of measure for weight? I have customers who enters weights in pounds (lb), not ounces.

I basically just added this. We run into an issue with the cubic meters though - do you know of an abstract way to do the conversions? For some reason I can't get the math to work by simply rooting the volume, converting the side, then squaring the converted side. I'll double check though, because I feel like that should work 😑

max3903 (approved):

lasley:
Doesn't look like we have much in the way of stakeholders in this repo, so I'm just merging on @max3903 approval & a decent amount of time sitting for any dissenters.

PR #19, opened 2017-11-10
[ADD] l10n_us_product_stock: Glue module 

lasley:
* Create glue module for when l10n_us_product and stock are installed. Overrides weight/volume in product view provided in stock

Depends:
- [x] #18

### DIFF
--- a/l10n_us_product_stock/__manifest__.py
+++ b/l10n_us_product_stock/__manifest__.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'USA - Stock',
+    'summary': 'This module overrides the product weight and volume to use '
+               'ounces and cubic inches. It automatically installs if you '
+               'have l10n_us_product and stock installed.',
+    'version': '10.0.1.0.0',
+    'category': 'Extra Tools',
+    'website': 'https://laslabs.com/',
+    'author': 'LasLabs, '
+              'Odoo Community Association (OCA)',
+    'license': 'LGPL-3',
+    'application': False,
+    'installable': True,
+    'auto_install': True,
+    'depends': [
+        'l10n_us_product',
+        'stock',
+    ],
+    'data': [
+        'views/product_template_view.xml',
+    ],
+}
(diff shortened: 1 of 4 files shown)


lasley:
Rebased now that #18 is merged. Ready for review

l10n_us_product_stock/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+              'Odoo Community Association (OCA)',
+    'license': 'LGPL-3',
+    'application': False,
+    'installable': True,
+    'auto_install': True,
+    'depends': [
+        'l10n_us_product',
+        'stock',
+    ],
+    'view': [
```
max3903: @lasley `data`?

max3903 (changes requested):

l10n_us_product_stock/__manifest__.py
lasley: For some reason there was a week that I decided I was going to do this in all the manifests I made. Literally a week long brain fart & this is a result 😆 🚀

l10n_us_product_stock/views/product_template_view.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_template_property_form" model="ir.ui.view">
+        <field name="name">product.template.weight.oz</field>
+            <field name="model">product.template</field>
+            <field name="inherit_id" ref="product.view_template_property_form"/>
```
max3903: @lasley Your view is not used. It should inherit from `stock.view_template_property_form`. That's where `group[@name='group_lots_and_weight']` is defined.

max3903 (changes requested):

max3903 (approved):

lasley:
Confirmed on Runbot too! Admittedly this is the first time I've run this module - up until now it was simply theory

image

lasley:
It's kind of interesting though because the base is actually KG so the OZ will always be slightly off. I wonder if I should note this in the readme? Example:

* oz input: 12345
  * oz saved: 12345.20
* oz input: 54321
  * oz saved: 54321.91

The cubic inches look fine because they are a significantly smaller scale than the base, so the math works better without high decimal precision.

lasley:
@obulkin please review when you get a chance

obulkin (approved):
LGTM. I would explain the rounding error issue in the ``l10n_us_product`` README as this is where the behavior actually gets introduced.

lasley:
Thanks @obulkin - note added in #22

Merging

PR #20, opened 2017-11-10
[ADD] l10n_us_delivery: Default stock uom to ounces

lasley:
* Create module to default stock picking UOMs to ounces, if available

### DIFF
--- a/l10n_us_delivery/README.rst
+++ b/l10n_us_delivery/README.rst
@@ -1,0 +1,51 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==============
+USA - Delivery
+==============
+
+This module makes ounces the default unit of measure for stock pickings.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/203/10.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/l10n-usa/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Dave Lasley <dave@laslabs.com>
+
+Do not contact contributors directly about support or help with technical issues.
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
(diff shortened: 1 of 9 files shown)


max3903:
@lasley I think you are missing places in Odoo interface where the unit of measure is shown:

* On the stock picking and stock move forms
* On the package details form
* On the delivery slip and picking operations reports

We made a module to use pounds (lb) by default instead of ounces with these views:

```
<?xml version="1.0"?>
<odoo>

    <!-- KG in the stock move form -->
    <record id="view_move_withweight_form" model="ir.ui.view">
        <field name="name">stock.move.form.weight</field>
        <field name="model">stock.move</field>
        <field name="inherit_id" ref="delivery.view_move_withweight_form"/>
        <field name="arch" type="xml">
            <field name="weight_uom_id" position="replace">
                <span> lb</span>
            </field>
        </field>
    </record>

    <!-- KG in the stock picking form -->
    <record id="view_picking_withcarrier_out_form" model="ir.ui.view">
        <field name="name">delivery.stock.picking_withcarrier.form.view</field>
        <field name="model">stock.picking</field>
        <field name="inherit_id" ref="delivery.view_picking_withcarrier_out_form"/>
        <field name="arch" type="xml">
            <field name="weight_uom_id" position="replace">
                <span> lb</span>
            </field>
        </field>
    </record>

    <!-- KG in the package details form -->
    <record id="view_quant_package_form_save" model="ir.ui.view">
        <field name="name">stock.quant.package.form.save</field>
        <field name="model">stock.quant.package</field>
        <field name="inherit_id" ref="delivery.view_quant_package_form_save"/>
        <field name="arch" type="xml">
            <xpath expr="//div[@class='o_row']/span" position="replace">
                <span>lb</span>
            </xpath>
        </field>
    </record>

    <!-- KG in the delivery slip report -->
    <record id="report_delivery_document2" model="ir.ui.view">
        <field name="name">report_delivery_document2</field>
        <field name="inherit_id" ref="delivery.report_delivery_document2"/>
        <field name="arch" type="xml">
            <span t-field="o.weight_uom_id" position="replace">
                <span>lb</span>
            </span>
        </field>
    </record>

    <!-- KG in the picking operations report -->
    <record id="report_shipping2" model="ir.ui.view">
        <field name="name">report_shipping2</field>
        <field name="inherit_id" ref="delivery.report_shipping2"/>
        <field name="arch" type="xml">
            <span t-field="o.weight_uom_id" position="replace">
                <span>lb</span>
            </span>
        </field>
    </record>
</odoo>
```

lasley:
In this instance there’s a proper uom that’s getting switched, so my assumption is that all these should just work (vs the product ones I just submitted). I’ll double check after I finish lunch. 

In hindsight also, it might make sense to only make these changes if the current user company is in America. What do you think? 


-Dave Lasley

lasley:
@max3903 - Thanks for the notes - I double checked everything:

* `delivery.view_quant_package_form_save` - Yup, hardcoded `kg`. I'll update that to display the UoM instead
* `delivery.view_move_withweight_form` - This one shouldn't need changing, the `weight_uom_id` is being directly displayed, which is being updated by the default
* `delivery.view_picking_withcarrier_out_form` - Same as above
* `delivery.report_delivery_document2` - Same as above
* `delivery.report_shipping2` - Same as above

Question - Is your above module also updating the `stock_picking.weight_uom_id` to the proper UoM (lb or oz)? If not, your users have been entering data as if they are using pounds, but the system is using `kg` still.

Issue #17, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Dependencies

https://github.com/OCA/account-financial-tools/issues/551
- [x] account_reversal

https://github.com/OCA/account-payment/issues/161
- [x] account_check_printing_report_base
- [x] account_payment_batch_process
- [x] account_payment_credit_card
- [x] partner_aging

https://github.com/OCA/account-invoice-reporting/issues/63
- [x] partner_time_to_pay

https://github.com/OCA/account-reconcile/issues/186
- [x] account_banking_reconciliation

# Modules to migrate

- [x] l10n_us_form_1099 by @max3903 in #21
- [ ] l10n_us_uom_profile
- [x] l10n_us_check_writing_address by @Sudarshan1607 in #25 and @murtuzasaleh in #31
- [x] l10n_us_account_profile by @Sudarshan1607 in #26 and @murtuzasaleh in #32

PR #21, opened 2017-12-05
[MIG] l10n_us_form_1099

max3903:
#17


l10n_us_form_1099/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "license": "AGPL-3",
+    "summary": "Add 1099 field to res.partner that will auto-check supplier",
+    "category": "Customers",
+    "maintainer": "Open Source Integrators",
+    "website": "https://www.opensourceintegrators.com",
+    "depends": ["contacts"],
+    "data": [
+        "views/res_partner.xml",
+    ],
+    "qweb": [
```
lasley: Please remove empty key

l10n_us_form_1099/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    is1099 = fields.Boolean('Is a 1099')
+    supplier = fields.Boolean(string='Is a Vendor',
```
lasley: Isn't this already a field on partner?

l10n_us_form_1099/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2017 Open Source Integrators <https://www.opensourceintegrators.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    is1099 = fields.Boolean('Is a 1099')
```
lasley: Please use snake case var names- `is_1099` & align the onchange `_on_change_is_1099`

l10n_us_form_1099/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+    supplier = fields.Boolean(string='Is a Vendor',
+                              help="Check this box if\
```
lasley: Please define one param per line as indicated in the multi-line fields declaration [example in the contribution](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models) (look in the code block slightly below the header I linked)

Additionally, the help string (if still multi-line), should break the string and continue again. Example:
```
x = fields.Char(
    help='one line'
         'another line',
)
```

lasley (changes requested):
Thanks @max3903. Am I missing something though? Should this module be doing anything other than just adding the field and performing the onchange? Where do you store the 1099?

lasley:
Also I was just looking at the commit history - any thoughts on squashing by author? IMO they don't provide much to the timeline in terms of notes or actual changes - most look like fixups that should have been squashed already.

gmader:
Hi all

We are not storing or even generating a 1099.  This is just a simple module
to flag a 1099 staffer.  You will still need to manage payment and IRS
reporting separately.

When companies hire others as contractors, the contractor in question may
work for themselves, another company or be the whole other company.  This
is a huge range, but it can be simplified down to “do I report their
payment to the IRS as a 1099 ?”

Does this help explain the module better?


Thanks
Greg

l10n_us_form_1099/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
...
+    @api.onchange('is_1099')
+    def _on_change_is_1099(self):
+
+        if self.is_1099 and not self.supplier:
+            self.supplier = True
+
+    @api.onchange('supplier')
+    def _on_change_supplier(self):
+
+        if self.is_1099 and not self.supplier:
```
lasley, line 21: I think it may make sense to also add a constraint to not allow a 1099 on a non-supplier?

lasley (commented):
Thanks for the clarification @gmader - this makes sense. One inline comment additional, although not a blocker.

We will need tests for the two methods though please @max3903

PR #141, opened 2025-02-19
[18.0][MIG] l10n_us_gaap

JasminSForgeFlow:
Standard Migration

Update Tests and add demo data for tests

@ForgeFlow


ThiagoMForgeFlow:
Why do you add a new demo company and remove the loading of accounts for the others? Doing that, the accounts in the template are only created for Demo Company.

JasminSForgeFlow:
> Why do you add a new demo company and remove the loading of accounts for the others? Doing that, the accounts in the template are only created for Demo Company.

In Odoo 18 try_loading works with demo data only, if we try to keep data file then below warning will come and tests are failing

`odoo.addons.account.models.chart_template: Incorrect usage of try_loading without a fully loaded registry. This could lead to issues. (YourCompany-us_gaap)`

ThiagoMForgeFlow:
> > Why do you add a new demo company and remove the loading of accounts for the others? Doing that, the accounts in the template are only created for Demo Company.
> 
> In Odoo 18 try_loading works with demo data only, if we try to keep data file then below warning will come and tests are failing
> 
> `odoo.addons.account.models.chart_template: Incorrect usage of try_loading without a fully loaded registry. This could lead to issues. (YourCompany-us_gaap)`

Have you tried passing this parameter in the try_loading function?
`<value name="install_demo" eval="True" />`

JasminSForgeFlow:
> > > Why do you add a new demo company and remove the loading of accounts for the others? Doing that, the accounts in the template are only created for Demo Company.
> > 
> > 
> > In Odoo 18 try_loading works with demo data only, if we try to keep data file then below warning will come and tests are failing
> > `odoo.addons.account.models.chart_template: Incorrect usage of try_loading without a fully loaded registry. This could lead to issues. (YourCompany-us_gaap)`
> 
> Have you tried passing this parameter in the try_loading function? `<value name="install_demo" eval="True" />`

Do you think that's technically correct? I take a look in Odoo standard addons but didn't find any example like this.

ayushin:
anybody's working on this?

jelenapoblet:
@ayushin I think we just need to do some reviews. @rrebollo can you help us check this migration ? thanks

l10n_us_gaap/data/demo_company.xml
rrebollo, line 1: Maybe it's time to fix the typo in the name of this file: "demo_company.xml". It should be fixed on the __manifest__ as well.

rrebollo (approved):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit.

ayushin:
by the way: no matter the installation sequence the accounting module installs the generic US chart of accounts instead of GAAP - something i miss?

ayushin:
There seem to be no way to install "us_gaap" template:  no matter the installation sequence the accounting module installs the generic US chart of accounts instead of GAAP 

the only workaround is to hard code / override "us_gaap" in the try_loading - even _guess_chart_template does not solve this

also the account groups are erroneously loaded as tax groups

rrebollo:
@JasminSForgeFlow can you clarify installation instruccions, according to @ayushin previous comment:

> There seem to be no way to install "us_gaap" template: no matter the installation sequence the accounting module installs the generic US chart of accounts instead of GAAP
> 
> the only workaround is to hard code / override "us_gaap" in the try_loading - even _guess_chart_template does not solve this
> 
> also the account groups are erroneously loaded as tax groups

ayushin:
yeah i am more worried about the account groups and also the gaap_mis_report is totally broken

JasminSForgeFlow:
> There seem to be no way to install "us_gaap" template: no matter the installation sequence the accounting module installs the generic US chart of accounts instead of GAAP
> 
> the only workaround is to hard code / override "us_gaap" in the try_loading - even _guess_chart_template does not solve this
> 
> also the account groups are erroneously loaded as tax groups

You can install it via Invoicing-> Configuration -> Settings and there from Fiscal localization package

image

you can select US Gaap option from it and click on save and it will install accounts and tax groups

image

image

ayushin:
@JasminSForgeFlow

1. When you install any accounting module it will automatically install the non-GAAP (generic) CoA and after that changing it has no effect
2. What is installed as tax groups actually account groups - these are not tax groups, you need to rename that file to account groups - it is complete nonsense as it is

3. The gaap_mis reports produces rubbish for PnL - needs work.

rrebollo:
@ayushin Could you confirm if you selected the United States as the database country during setup? If so, Odoo automatically installs a default localization package for US accounting (fiscal rules) when the Accounting app is installed, which might explain some of your issues.

> * When you install any accounting module it will automatically install the non-GAAP (generic) CoA and after that changing it has no effect

l10n_us_gaap/readme/USAGE.md
```diff
@@ -0,0 +1,14 @@
...
+
+Companies should review and alter the sample CoA to fulfill the specific
+needs of their business.
+
+U.S. SEC not allow the classification of expenses by nature, thus the
+classification is proposed by function.
+
+This module contains in the 'docs' folder a a sample Trial Balance
+generated using the [OCA Account Financial Report
+module](https://github.com/OCA/account-financial-reporting/blob/17.0/account_financial_report).
```
Daniel15: Link needs to be updated to 18.0

edescalona:
Hi @JasminSForgeFlow , could you please correct the change requests?

jelenapoblet:
This seems to be stale since January

l10n_us_gaap/data/demo_company.xml
JasminSForgeFlow: Fixed, Thanks

l10n_us_gaap/readme/USAGE.md
JasminSForgeFlow: Fixed, Thanks

edescalona (approved):
LGTM

PR #158, opened 2025-09-14
[17.0][MIG] account banking ach direct debit

crrodrigueztrujillo:
@BinhexTeam

Migration from Plaid 16 to 17.


antoniodavid (changes requested):
Thanks for your contribution. You should squash the last two commit in one and fix the tests cases.

PR #155, opened 2025-08-14
[18.0][FIX] account_banking_ach_credit_transfer: fix tests

amh-mw:
The due_date field in the account.payment.line.create wizard was renamed to filter_date upstream in bank-payment/account_payment_order.

PR #163, opened 2025-10-27
[18.0][FIX] account_banking_ach_credit_transfer: adapt test to new field

JordiMForgeFlow:
Needed due to https://github.com/OCA/bank-payment/pull/1487

CC @JordiBForgeFlow @jelenapoblet

### DIFF
--- a/account_banking_ach_credit_transfer/tests/test_account_banking_ach_credit_transfer.py
+++ b/account_banking_ach_credit_transfer/tests/test_account_banking_ach_credit_transfer.py
@@ -64,7 +64,7 @@ class TestACHCreditTransfer(BaseCommon):
             )
         )
         line_create_form.date_type = "due"
-        line_create_form.due_date = fields.Date.today()
+        line_create_form.filter_date = fields.Date.today()
         line_create_form.payment_mode = "any"
         line_create_form.target_move = "all"
         line_created_due = line_create_form.save()


JordiMForgeFlow:
@adasatorres this PR is aimed at fixing the error from the account_banking_ach_credit_transfer test.

However, it now fails due to a deprecation warning of the usage of pkg_resources inside the stdnum package.

The issue was already fixed in python-stdnum library (https://github.com/arthurdejong/python-stdnum/commit/7761e42764b4730335943b7b43779446ad459db4) but Odoo pins an older version where the fix is not there (python-stdnum==1.19 ; python_version >= '3.11')

Not sure what should be the resolution here: should we pin a newer version for the module or should we allow the warning on the CI somehow?

adasatorres:
Hi @JordiMForgeFlow, 

My suggestion, add the fixed version as an external dependency to ensure it isn't installed with another version, thus forcing a manual update. Additionally, consider opening an issue or PR in the Odoo CE repository to inform the Odoo maintainers about this problem, since anyone using this library at version 1.19 would likely encounter the same error.

I don't believe temporarily suppressing the warning is the right approach.

Keep in mind that the runboat and the tests run in a container with Python 3.10 and python-stdnum==1.17.

nilshamerlinck:
superseded by https://github.com/OCA/l10n-usa/pull/155

PR #173, opened 2026-05-20
[18.0][ADD] l10n_us_sales_tax_engine: Hybrid US Sales Tax Engine

crrodrigueztrujillo:
## Summary

This PR adds `l10n_us_sales_tax_engine`, a provider-agnostic hybrid sales tax
calculation engine for all 50 US states, developed by **@BinhexTeam**.

- Local tax rate database (Florida DOR seed data — 67 counties) queried first
  for zero-latency lookups
- External API fallback chain: **ZipTax → API Ninjas → TaxJar**
  (priority-ordered, skips providers that have exhausted their monthly limit)
- Response cache with configurable TTL (default 30 days) to minimize API calls
- Full **immutable audit log** per calculation (source, rate breakdown,
  provider, address)
- **Nexus management** per company/state (economic nexus threshold support)
- **Product fiscal categories**: TANGIBLE, SERVICE, FOOD, MEDICINE, DIGITAL,
  SHIPPING, EXEMPT
- Sale Order and Customer Invoice integration — auto-calculates on confirm
- Internal REST endpoints (`/api/v1/us_tax/calculate`,
  `/api/v1/us_tax/rate_lookup`)
- Provider-agnostic architecture — new providers can be added without touching
  the engine core

## Test plan
- [ ] Enable engine in *Accounting → Configuration → Settings*
- [ ] Configure nexus states
- [ ] Sale Order with US address → **Calculate US Tax** → verify Calculation Log
- [ ] Fallback chain: disable local provider → confirm API provider used
- [ ] Cache: second calculation for same ZIP → `source = cache`
- [ ] Exempt rule: NY services → `source = exempt_rule`, tax = 0
- [ ] Import Florida DOR CSV via *Configuration → Import Rates*

## Screenshots

11 screenshots included in `static/description/` showing settings, menu,
nexus, calculation logs, providers and tax rates.

---
*Contribution by Carlos R. Rodriguez — [@BinhexTeam](https://github.com/BinhexTeam)*


rrebollo:
@crrodrigueztrujillo please squash your commits into the first one.

rrebollo:
@crrodrigueztrujillo please prefix the PR name with the targeted version, like "[18.0] [ADD] ..."

l10n_us_sales_tax_engine/__manifest__.py
```diff
@@ -0,0 +1,47 @@
...
+    "website": "https://github.com/OCA/l10n-usa",
+    "license": "LGPL-3",
+    "development_status": "Alpha",
+    "depends": [
+        "account",
+        "sale_management",
+        "mail",
+    ],
+    "external_dependencies": {
+        "python": ["requests"],
```
rrebollo: Don't just take my word for it, but 'requests' is a core dependency listed in the requirements.txt for 18.0 CE, so it's always available. I don't think it's necessary to explicitly mention it here.

l10n_us_sales_tax_engine/setup.cfg
rrebollo, line 1: I think this shouldn't be part of the migration — I haven't seen it before in OCA addons.

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
```diff
@@ -0,0 +1,120 @@
...
+        rates,
+        ttl_hours,
+        request_payload=None,
+        response_payload=None,
+        zip_code=None,
+        state_id=None,
+    ):
+        """Upsert a cache entry."""
+        now = fields.Datetime.now()
+        expires = now + timedelta(hours=ttl_hours) if ttl_hours else False
```
rrebollo: Date/Datetime built-in helpers are great, like https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.fields.Datetime.add

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
```diff
@@ -0,0 +1,120 @@
...
+            "city_rate": rates.get("city_rate", 0.0),
+            "district_rate": rates.get("district_rate", 0.0),
+            "source_date": rates.get("source_date"),
+            "cached_at": now,
+            "expires_at": expires,
+            "request_payload": json.dumps(request_payload or {}),
+            "response_payload": json.dumps(response_payload or {}),
+        }
+        if existing:
+            existing.write(vals)
```
rrebollo:
```suggestion
            existing.update(vals)
```

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        return self.create(vals)
+
+    @api.model
+    def purge_expired(self):
+        """Called by cron to archive expired entries."""
+        expired = self.search(
+            [
+                ("expires_at", "<", fields.Datetime.now()),
+                ("active", "=", True),
```
rrebollo:
```suggestion
```
Default behavior is to look for active records.

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
```diff
@@ -0,0 +1,120 @@
...
+            ]
+        )
+        expired.write({"active": False})
```
rrebollo:
```suggestion
        expired.active = False
```

l10n_us_sales_tax_engine/models/us_tax_zip_mapping.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2026 Binhex - Carlos R. Rodriguez.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+
+
+class UsTaxZipMapping(models.Model):
+    _name = "us.tax.zip.mapping"
+    _description = "US Tax ZIP to Jurisdiction Mapping"
+    _order = "zip, confidence desc"
```
rrebollo, line 9: Define _rec_name or display_name field.

l10n_us_sales_tax_engine/models/us_tax_import_batch.py
rrebollo, line 1: My comment applies to the entire import mechanism. Rather than building another importer from scratch, I would prefer to see an approach that extends [OCA's connector_importer](https://github.com/OCA/connector-interfaces/tree/18.0/connector_importer).

l10n_us_sales_tax_engine/models/us_tax_rate.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Preference: specific product category > generic (null category)
+        """
+        domain = [
+            ("jurisdiction_id", "=", jurisdiction_id),
+            ("effective_date", "<=", date),
+            "|",
+            ("end_date", "=", False),
+            ("end_date", ">=", date),
+            ("active", "=", True),
```
rrebollo:
```suggestion
            ("active", "=", True),
            ("product_tax_category_id", "=", product_category_id)
```

l10n_us_sales_tax_engine/models/us_tax_rate.py
```diff
@@ -0,0 +1,94 @@
...
+        if product_category_id:
+            # Try specific category first
+            specific = self.search(
+                domain + [("product_tax_category_id", "=", product_category_id)],
+                order="effective_date desc",
+                limit=1,
+            )
+            if specific:
+                return specific
+        # Fallback to generic rate (no category)
```
rrebollo:
```suggestion    
```

l10n_us_sales_tax_engine/models/us_tax_rate.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.search(
+            domain + [("product_tax_category_id", "=", False)],
```
rrebollo:
```suggestion
            domain,
```

rrebollo (commented):
First of all, thank you for the effort and value of this contribution to the OCA ecosystem, and especially for the US localization space.

I tried to perform a thorough review, but with more than 5k added lines it becomes very difficult to properly assess all aspects of the implementation. With that in mind, I wanted to share a couple of broader considerations beyond implementation details:

* Splitting the solution into multiple addons could help significantly from a review and maintainability perspective by reducing the volume and scope of each component. Whether this also brings functional benefits in terms of avoiding unnecessary installations is something I cannot properly judge from my side.

* I would also encourage considering reuse of patterns and solutions that the community has already adopted for common problems, such as `connector`, `connector_importer`, `component`, or `queue_job`. From my own experience, I know these frameworks are not always well documented and require a certain technical background, so sometimes reimplementing things from scratch feels simpler and faster. However, over time, solutions that initially seemed sufficient can start growing organically until they become difficult to maintain or extend.

I would be interested in hearing your thoughts on these points.

Once again, thank you for the contribution.

usamahassan-rt:
What about california ?? CDFTA apis are best source ?

l10n_us_sales_tax_engine/__manifest__.py
crrodrigueztrujillo: Good catch! Removed — `requests` is indeed a core Odoo 18 CE dependency.

l10n_us_sales_tax_engine/setup.cfg
crrodrigueztrujillo: Removed. Thanks for the heads up.

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
crrodrigueztrujillo: Updated to use `fields.Datetime.add()` as suggested. Thanks!

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
crrodrigueztrujillo: Applied. Changed to `update()`.

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
crrodrigueztrujillo: Agreed — removed. Odoo already handles active records by default.

l10n_us_sales_tax_engine/models/us_tax_api_cache.py
crrodrigueztrujillo: Applied. Updated to `expired.active = False`.

l10n_us_sales_tax_engine/models/us_tax_zip_mapping.py
crrodrigueztrujillo: Added `_rec_name = "zip"`. Thanks!

l10n_us_sales_tax_engine/models/us_tax_import_batch.py
crrodrigueztrujillo: We appreciate the suggestion. The import mechanism here is designed for loading government-published flat files (Florida DOR, future state CSVs) on an ad-hoc basis — not a recurring sync flow. Adding `connector_importer` as a dependency would significantly increase the installation footprint for what is essentially a periodic manual operation. We'd prefer to keep it lightweight in v1 and revisit in a follow-up if the community agrees it's the right direction.

l10n_us_sales_tax_engine/models/us_tax_rate.py
crrodrigueztrujillo: Applied.

l10n_us_sales_tax_engine/models/us_tax_rate.py
crrodrigueztrujillo: Applied.

l10n_us_sales_tax_engine/models/us_tax_rate.py
crrodrigueztrujillo: Applied. Also updated the test fixtures to align with the simplified lookup logic.

crrodrigueztrujillo:
Done! All commits squashed into one.

crrodrigueztrujillo:
Done! Updated to `[18.0][ADD] l10n_us_sales_tax_engine: Hybrid US Sales Tax Engine`.

crrodrigueztrujillo:
@usamahassan-rt Great point! California's CDTFA is one of the most complex tax jurisdictions in the US. The architecture is provider-agnostic — adding a CDTFA-specific provider only requires implementing a new class extending `ProviderBase`. Contributions are very welcome!

crrodrigueztrujillo:
Thank you for the thorough review @rrebollo and for taking the time despite the volume of changes.

Regarding splitting into multiple addons — we agree it's worth considering as the module grows and we'll evaluate it for a follow-up PR.

On the OCA frameworks (`connector`, `component`, `queue_job`): the tax rate import here is designed for loading government-published flat files (Florida DOR, future state CSVs) on an ad-hoc basis — not a recurring sync flow. Adding `connector_importer` as a dependency would significantly increase the installation footprint for what is essentially a periodic manual operation. We'd prefer to keep it lightweight in v1 and revisit in a follow-up if the community sees value in it.

All other inline suggestions have been applied — thank you for the detailed feedback!

rrebollo (approved):
Code Review. LGTM!

jelenapoblet:
> @usamahassan-rt Great point! California's CDTFA is one of the most complex tax jurisdictions in the US. The architecture is provider-agnostic — adding a CDTFA-specific provider only requires implementing a new class extending `ProviderBase`. Contributions are very welcome!

Following @rrebollo recommendation, maybe we could have a sale_tax_engine_base and then create a sale_tax_provider_xxx for each one of them, this way we could allow more contributions to this idea. What do you think ?

crrodrigueztrujillo:
@jelenapoblet Thank you for the suggestion — we really like the idea of a `l10n_us_sales_tax_engine_base` + `l10n_us_sales_tax_provider_xxx` architecture. It aligns well with the provider-agnostic design we already have internally, and would make community contributions much easier.

We're open to restructuring the module this way before merging if that's the direction the OCA community prefers. Could you and @rrebollo confirm if this is a requirement for approval, or a recommendation for a future iteration? That will help us prioritize the refactor accordingly.

rrebollo:
@crrodrigueztrujillo, I don't have merge superpowers — I just made a review. Your contribution already shows a lot of work, and I'm not here to make you do even more.

That said, in my humble opinion, the distributed architecture is much cleaner.

jelenapoblet:
I think we're alligned @rrebollo , this way we'll have a better architecture and more contributors will be able to participate.

crrodrigueztrujillo:
Thanks @jelenapoblet and @rrebollo — we're aligned. We'll restructure into a base
engine + one addon per provider, which makes community contributions much easier.

I noticed @ledoent has already started moving in this direction (the standalone TaxJar
provider addon and the extensions in #183), which is great to see. To avoid duplicated
effort and keep naming consistent across the ecosystem, a couple of questions:

1. Naming: should we standardize on `l10n_us_sales_tax_provider_<name>` for the provider
   addons (the form already used in @ledoent's TaxJar PR)? Happy to align with that
   convention so all providers follow the same pattern.

2. Coordination: we have ZipTax, API Ninjas and TaxJar providers ready to split out from
   our engine. @ledoent, since you've started a TaxJar provider addon — would you prefer
   we focus on ZipTax/API Ninjas and you keep TaxJar, or should we coordinate differently?

Our plan: keep this PR (#173) as the base engine, remove the hardcoded provider imports
in favor of a lightweight Python registry so providers self-register, and submit each
provider as its own follow-up addon depending on the base. Does that sequencing work for
the maintainers?

jelenapoblet:
@crrodrigueztrujillo let's move forward to build the base and both ZipTax and API Ninja, we can deal with TaxJar once we make sure #183 works with #173

crrodriguezt:
@rrebollo @jelenapoblet — Updated this PR following our earlier discussion.

Changes since your review:
- Restructured as a provider-agnostic base engine: removed the hardcoded
  ZipTax, API Ninjas, and TaxJar provider classes. Providers now
  self-register via a registry extension point
  (`_provider_service_classes()` on `us.tax.provider`) — the base never
  imports a concrete provider.
- ZipTax and API Ninjas now ship as their own follow-up addons (opening
  those PRs right after this one is confirmed green).
- TaxJar is intentionally not included — leaving that to #183.
- A few bug fixes found during manual QA: background calculation no
  longer requires the US Tax groups for any user, exempt/0% lines now
  get an explicit tax record instead of an empty one, REST API body
  parsing, and a rate-fallback fix for category-less rates.

Given the scope of the restructuring, re-requesting review on this
version rather than assuming the earlier approval still applies.

crrodrigueztrujillo:
Fixed a leftover bug in \`pyproject.toml\`: the \`[build-system]\` block had a stale, invalid \`build-backend = "setuptools.backends.legacy:build"\` (pre-dating the \`whool-init\` hook, which only adds a \`[build-system]\` block if none exists — it never validates an existing one). Corrected to the standard whool values (\`requires = ["whool"]\`, \`build-backend = "whool.buildapi"\`). This surfaced via #196's temporary \`test-requirements.txt\` reference to this PR — pip couldn't build this addon as a package at all with the broken backend. Re-pushed (amended, same commit).

jelenapoblet:
Thanks @crrodrigueztrujillo , this is the way!

l10n_us_sales_tax_engine/importers/importer_florida_dor.py
rrebollo, line 1: Why is this still part of the core engine addon? If the same criteria used to justify keeping it here were applied to other states, this could easily grow over time and become too large to maintain as a single piece.

l10n_us_sales_tax_engine/controllers/us_tax_api.py
```diff
@@ -0,0 +1,142 @@
+# Copyright 2026 Binhex - Carlos R. Rodriguez.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+import logging
+from datetime import date
```
rrebollo: Why not use fields.Date and its helper methods?

l10n_us_sales_tax_engine/importers/importer_florida_dor.py
rrebollo, line 1: Why is this still part of the core engine addon? If the same criteria used to justify keeping it here were applied to other states, this could easily grow over time and become too large to maintain as a single piece.

l10n_us_sales_tax_engine/importers/importer_generic_csv.py
rrebollo, line 1: I would explore Odoo's built-in CSV import functionality and see if there's any component we can reuse, rather than building everything from scratch. I understand you'd have more control, but I don't think it would be that complex to hook into a specific point of Odoo's import mechanism, allowing you to reuse some of its benefits while still having the freedom to customize.

The idea here is to outsource maintenance, right? Also, if you rely on a solid Odoo foundation, you won't constantly have to fix issues related to basic import flows—the particularities of the files being imported are what you'd really need to handle. So I'd explore whether there's a manageable pathway through that approach.

l10n_us_sales_tax_engine/models/account_move.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+        except Exception as exc:
+            _logger.error("US Tax calculation error on invoice %s: %s", self.name, exc)
+            raise
+
+    def action_post(self):
+        """Auto-calculate tax on posting if engine is active."""
+        ICP = self.env["ir.config_parameter"].sudo()
+        if ICP.get_param("l10n_us_tax.engine_active", "False") == "True":
+            for move in self.filtered(lambda m: m.move_type == "out_invoice"):
```
rrebollo: what about rufunds/credit notes?

l10n_us_sales_tax_engine/models/account_move.py
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    "us_tax_source": result.get("source", ""),
+                    "us_tax_calculated_at": fields.Datetime.now(),
+                }
+            )
+        except Exception as exc:
+            _logger.error("US Tax calculation error on invoice %s: %s", self.name, exc)
+            raise
+
+    def action_post(self):
```
rrebollo: I notice that when extending the posting behavior of account moves, the usual approach is to override the `_post` method, not `action_post`. The `_post` method performs the actual posting of the account move, while `action_post` typically handles some validations around the posting process.

I'd ask you to consider this and review your implementation, because often the modifications are made in the wrong place. Additionally, the parent `action_post` method can, in some paths, trigger a wizard. If you check the base code in Odoo 18, you'll see it. This means that your tax calculation modification would depend on whether the user follows the "happy path" through the wizard flow, which would occur after your process, and doesn't enter any alternative flow.

Given that scenario, I think it would most likely be better to override the `_post` method instead of `action_post`. Please consider this and review — I could be wrong.

https://github.com/odoo/odoo/blob/f80fb6033e13221595fa8f097584b6fcd9c9230c/addons/account/models/account_move.py#L5556

l10n_us_sales_tax_engine/models/res_config_settings.py
```diff
@@ -0,0 +1,56 @@
...
+        default=0.7,
+        help="Minimum ZIP-to-jurisdiction confidence to use local data (0.0–1.0).",
+    )
+
+    def _sync_provider_active(self, code, enabled):
+        provider = (
+            self.env["us.tax.provider"].sudo().search([("code", "=", code)], limit=1)
+        )
+        if provider:
+            provider.write({"active": bool(enabled)})
```
rrebollo:
```suggestion
            provider.active = bool(enabled)
```

l10n_us_sales_tax_engine/models/sale_order.py
```diff
@@ -0,0 +1,58 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    us_tax_source = fields.Char(
+        string="Tax Source",
+        readonly=True,
```
rrebollo: When you make a field `readonly=True` at the model level, it isn't even available for import. That's why OCA often recommends considering making fields readonly only at the view layer instead. This recommendation applies to all model fields defined with `readonly=True`. It's not a hard rule, but you might want to evaluate each case individually.

l10n_us_sales_tax_engine/models/sale_order.py
```diff
@@ -0,0 +1,58 @@
...
+        help="Source used for the last US tax calculation.",
+    )
+    us_tax_calculated_at = fields.Datetime(
+        string="Tax Calculated At",
+        readonly=True,
+    )
+    us_tax_total = fields.Monetary(
+        string="US Tax Total",
+        compute="_compute_us_tax_total",
+        store=False,
```
rrebollo:
```suggestion
```

l10n_us_sales_tax_engine/models/sale_order.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    @api.depends("tax_totals", "amount_tax")
+    def _compute_us_tax_total(self):
+        for order in self:
+            order.us_tax_total = order.amount_tax
```
rrebollo: If the relation is so straightforward, why would you need another field?

l10n_us_sales_tax_engine/models/us_tax_calculation_log.py
```diff
@@ -0,0 +1,120 @@
...
+            if rec.res_model and rec.res_id:
+                try:
+                    doc = self.env[rec.res_model].browse(rec.res_id)
+                    rec.document_ref = (
+                        doc.display_name or f"{rec.res_model},{rec.res_id}"
+                    )
+                except Exception:
+                    rec.document_ref = f"{rec.res_model},{rec.res_id}"
+            else:
+                rec.document_ref = ""
```
rrebollo: Refactor this so you don't have to call browse() for every single record, but rather work in batch.

l10n_us_sales_tax_engine/models/us_tax_nexus.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @api.model
+    def has_nexus(self, company_id, state_id):
+        """Check if a company has active nexus in a given state."""
+        return bool(
+            self.search(
+                [
+                    ("company_id", "=", company_id),
+                    ("state_id", "=", state_id),
+                    ("active", "=", True),
```
rrebollo:
```suggestion
```

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    limit_status = fields.Selection(
+        [
+            ("ok", "OK"),
+            ("warning", "Warning (>80%)"),
+            ("exceeded", "Limit Exceeded"),
+            ("unlimited", "Unlimited"),
+        ],
+        compute="_compute_limit_status",
+        store=False,
```
rrebollo:
```suggestion
```

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    limit_usage_pct = fields.Float(
+        compute="_compute_limit_status",
+        store=False,
```
rrebollo:
```suggestion
```

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+                if pct >= 1.0:
+                    rec.limit_status = "exceeded"
+                elif pct >= LIMIT_WARNING_THRESHOLD:
+                    rec.limit_status = "warning"
+                else:
+                    rec.limit_status = "ok"
+
+    # ── Limit checks ──────────────────────────────────────────────────────────
+
+    def is_limit_reached(self) -> bool:
```
rrebollo: For the sake of consistency, you shouldn't use type hints here.

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+        """Return True if this provider has reached its monthly call limit."""
+        self.ensure_one()
+        if self.monthly_call_limit == 0:
+            return False  # 0 = unlimited
+        return self.calls_this_month >= self.monthly_call_limit
+
+    def is_limit_warning(self) -> bool:
```
rrebollo: For the sake of consistency, you shouldn't use type hints here.

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+            return False
+        return self.calls_this_month >= (
+            self.monthly_call_limit * LIMIT_WARNING_THRESHOLD
+        )
+
+    def increment_call_counter(self):
+        """Increment monthly call counter and send alert if threshold reached."""
+        self.ensure_one()
+        new_count = self.calls_this_month + 1
+        self.sudo().write({"calls_this_month": new_count})
```
rrebollo:
```suggestion
        self.sudo().calls_this_month = new_count
```

l10n_us_sales_tax_engine/models/us_tax_provider.py
```diff
@@ -0,0 +1,239 @@
...
+
+        if self.monthly_call_limit > 0:
+            pct = new_count / self.monthly_call_limit
+            # Send 80% warning once per month
+            if pct >= LIMIT_WARNING_THRESHOLD and not self.limit_alert_sent:
+                self.sudo().write({"limit_alert_sent": True})
```
rrebollo: Same pattern here. You're setting a single field, so use the ORM accordingly and consistently across the codebase.

l10n_us_sales_tax_engine/security/us_tax_security.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
rrebollo:
```suggestion
<odoo noupdate="1">
```

l10n_us_sales_tax_engine/services/tax_engine.py
```diff
@@ -0,0 +1,639 @@
...
+            if source in ("exempt_rule", "exempt_nexus"):
+                exempt_tax = self._get_or_create_exempt_tax(order.company_id)
+                line.write(
+                    {"tax_id": [(6, 0, [exempt_tax.id])] if exempt_tax else [(5, 0, 0)]}
+                )
+                continue
+
+            tax = self._get_or_create_state_tax(state_code, order.company_id, rate)
+            if tax:
+                line.write({"tax_id": [(6, 0, [tax.id])]})
```
rrebollo: You don't need the write() call, and https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.fields.Command

l10n_us_sales_tax_engine/services/tax_engine.py
```diff
@@ -0,0 +1,639 @@
...
+                    "company_id": company.id,
+                    "tax_group_id": tax_group.id,
+                    "description": f"US Sales Tax — {state_code} @ {amount_pct:.4g}%",
+                }
+            )
+            _logger.info('Created tax "%s" at %s%%', name, amount_pct)
+        elif not tax.tax_group_id or tax.tax_group_id.name != "US Sales Tax":
+            tax_group = TaxGroup.search([("name", "=", "US Sales Tax")], limit=1)
+            if tax_group:
+                tax.write({"tax_group_id": tax_group.id})
```
rrebollo: write again?

l10n_us_sales_tax_engine/wizards/us_tax_import_wizard.py
```diff
@@ -0,0 +1,108 @@
...
+        domain=[("country_id.code", "=", "US")],
+        required=True,
+    )
+    effective_date = fields.Date(required=True, default=fields.Date.today)
+    file_data = fields.Binary(string="File", required=True)
+    file_name = fields.Char()
+    delimiter = fields.Char(default=",", size=1)
+    preview_html = fields.Html(readonly=True)
+
+    def action_preview(self):
```
rrebollo, line 38: This feature reinforces my thoughts about the possibility of forking the built‑in import functionality.

l10n_us_sales_tax_engine/pyproject.toml
rrebollo, line 1: For me, this file is normally generated by the pre-commit hooks at the repository level. If you have any issues with it, you should check that your local tooling is aligned with the GitHub deployment.

rrebollo (changes requested):
Overall, the development is maturing. There are a number of what I would call bad practices or patterns distributed throughout the addon's codebase — I only pointed them out in specific places, but they should be corrected across the board.

In particular, there's excessive use of direct calls to the `write` method on recordsets. The problem with this is that it prevents the ORM from applying certain optimizations when you explicitly call `write`. The ORM generally knows when it's more convenient to handle caching and overall performance improvements. Specifically, when you have only one record in the recordset, an explicit direct `write` call is discouraged. Even when dealing with multiple recordsets, I often prefer alternatives like the `update` method.

There are also some other details that aren't necessarily bad practices but relate to code consistency and design. I'd like to point out that my observations shouldn't be corrected only where I flagged them, but you should look for other occurrences throughout the code.

I think smoothing out these issues — along with a few design considerations that don't necessarily require correction but rather some self-reflection — would be beneficial. Sometimes I ask why something wasn't done in a certain way. It might be perfectly legitimate to do it the way it's currently done, but it's good to leave the justification there for anyone who might review it later. An explicit note addressing any doubts users might have would also be helpful.

Overall, I believe that fixing these points would make the PR much more ready for review. From a design perspective, I also noticed that the tax providers for Texas and Florida were left in place again, as I saw in part of the core. The motivation there might be pragmatic, and that's fine. I left a note suggesting we rethink it, but as I said, these are all just suggestions. Ultimately, as the maintainer or developer of the PR, you can take them or leave them — I'm putting them out there.

crrodriguezt:
Thanks for the detailed pass, @rrebollo — applying these across the board, not just where flagged (single-record `write()` → direct assignment in `account_move.py`, `sale_order.py`, `us_tax_provider.py`, `tax_engine.py` (using `fields.Command` for the x2many taxes), `us_tax_import_wizard.py`; `noupdate="1"` on `us_tax_security.xml`; removed the unused `us_tax_total` field; batched `_compute_document_ref` by `res_model` instead of `browse()` per record; moved `readonly` to the view layer for `us_tax_source`/`us_tax_calculated_at`).

One correction while applying these — the `action_post()` → `_post()` migration surfaced a real bug in the original code: computing tax *after* `super()._post()` hits "you cannot modify the taxes related to a posted journal item" once the move is actually posted, since lines are no longer writable at that point. Fixed by computing tax *before* calling `super()._post()` — reaching `_post()` at all already guarantees the move will post (no wizard diversion happens below `action_post()`), so this is safe. Also extended scope to `out_refund` to match what the manual "Calculate US Tax" button already allowed. Added `tests/test_account_move.py` (3 new tests) since this exact path had zero coverage before.

On the type hints: kept them in `services/` (plain Python classes, not Odoo models) where they're used consistently across all 3 modules — only removed the 2 on `us_tax_provider.py` model methods, which seems to be what you meant given the rest of the repo's Odoo models don't use them either.

crrodriguezt:
On `pyproject.toml`: I want to flag a correction here, since it's relevant to how we (and others) should treat this file going forward.

I checked the actual source of the pinned hook (`sbidoul/whool` @ `v1.2`, same revision used locally and in this repo's CI — `.pre-commit-config.yaml`). `whool-init`'s logic only *creates* the `[build-system]` block if the file has none at all:

```python
if not pyproject_toml_path.exists():
    ... # create from scratch
else:
    pyproject_toml = tomllib.load(f)
    if "build-system" not in pyproject_toml:
        ... # add one
    # if [build-system] already exists (even with bad content), it's left untouched
```

So it's true that pre-commit *generates* this file for new addons (confirmed: `l10n_us_sales_tax_provider_ziptax` and `l10n_us_sales_tax_provider_api_ninjas` both got the correct `whool` values automatically, since they started with no `pyproject.toml` at all). But it never *validates* or re-generates an *existing* block — which is exactly the situation here: this addon's `pyproject.toml` had a stale `build-backend = "setuptools.backends.legacy:build"` from before `whool` was adopted on this module, and the hook silently left it alone on every run (locally and in CI, same pinned version, so it wasn't a tooling-drift issue either). It only became visible when #196 tried to `pip install` this addon as a package via the OCA temporary-cross-PR-dependency mechanism — `odoo-bin` itself never reads `pyproject.toml`, so the bug was invisible until something actually tried to build a wheel from it. The fix I pushed matches `whool`'s own generated values exactly, so it's the canonical state, not a workaround.

crrodrigueztrujillo:
Also applied: `controllers/us_tax_api.py` now uses `fields.Date.today()` instead of `datetime.date.today()` (2 call sites), and `us_tax_nexus.py::has_nexus()` dropped the redundant `("active", "=", True)` domain term (same pattern as the `us_tax_api_cache.py` fix earlier — `search()` already defaults to active records).

l10n_us_sales_tax_engine/models/account_move.py
crrodrigueztrujillo: Good catch — fixed. `_post()` now runs for both `out_invoice` and `out_refund` (matches the scope the manual "Calculate US Tax" button already allowed in the view). Confirmed with a new test, `test_tax_auto_calculated_on_credit_note_post` in `tests/test_account_move.py`.

l10n_us_sales_tax_engine/models/account_move.py
crrodrigueztrujillo: You were right, and it surfaced a real bug while applying it: the original code computed tax *after* `super().action_post()`/`_post()` would have hit "you cannot modify the taxes related to a posted journal item" once the move actually posts, since lines stop being writable at that point. Fixed by overriding `_post(self, soft=True)` and computing tax *before* calling `super()._post()` — reaching `_post()` at all already guarantees the move will post (no wizard diversion below that point), so it's safe to compute there. Added `tests/test_account_move.py` (3 cases) since this path had zero coverage before. Thanks for pushing on this.

l10n_us_sales_tax_engine/importers/importer_florida_dor.py
crrodrigueztrujillo: I documented the reasoning for this directly in `importers/importer_base.py` (class docstring) rather than just here, so it doesn't get lost: this is ad-hoc, irregularly-shaped government file imports on a manual/periodic basis (no recurring sync to reconcile against), and the actual work is a handful of domain-specific upserts (jurisdiction → rate → ZIP mapping), not a generic field-mapping problem. Happy to revisit if a recurring multi-state sync need comes up — for now I don't think the added dependency footprint is justified. Open to discussing further if you still see it differently.

crrodrigueztrujillo: Same reasoning as the other thread on this file — documented now in `importers/importer_base.py`'s class docstring so it's not just buried in a PR comment.

l10n_us_sales_tax_engine/importers/importer_generic_csv.py
crrodrigueztrujillo: Same design note applies here — added to `importers/importer_base.py`'s docstring: there's no model whose fields map 1:1 to a DOR/generic export row, so Odoo's built-in import (column→field mapping) doesn't fit the shape of this problem either. Open to revisiting if the scope grows beyond "upload a file a few times a year."

l10n_us_sales_tax_engine/wizards/us_tax_import_wizard.py
crrodrigueztrujillo: Understood — same architecture discussion as the importer threads above. Documented the reasoning in `importers/importer_base.py`'s docstring rather than re-litigating it per-file. Let's keep tracking this as a possible future iteration rather than blocking this PR on it, unless you feel strongly it should gate the merge.

l10n_us_sales_tax_engine/pyproject.toml
crrodrigueztrujillo: Checked the actual source of the pinned hook (sbidoul/whool v1.2, same revision used here and locally): `whool-init` only *creates* the `[build-system]` block if the file has none at all — it never validates or regenerates an *existing* block, even a broken one. That's exactly this case: the file had a stale `build-backend` from before this addon adopted whool, and the hook silently left it alone on every run. It only became visible when #196 tried to pip-install this addon via the cross-PR dependency mechanism. The fix I pushed matches whool's own generated values exactly, so it's the canonical state, not a workaround — full details in my PR comment above if useful.

l10n_us_sales_tax_engine/models/us_tax_calculation_log.py
```diff
@@ -0,0 +1,122 @@
...
+            if rec.res_model and rec.res_id:
+                recs_by_model.setdefault(rec.res_model, []).append(rec)
+
+        for res_model, recs in recs_by_model.items():
+            existing = self.env[res_model].browse([r.res_id for r in recs]).exists()
+            names = {doc.id: doc.display_name for doc in existing}
+            for rec in recs:
+                rec.document_ref = names.get(rec.res_id) or (
+                    f"{rec.res_model},{rec.res_id}"
+                )
```
rrebollo, line 103: Better. I'm guessing your implementation is more performant, but if you want something more readable, take a look at the [`grouped`](https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.models.Model.grouped) method. It lets you organize your recordset into a dictionary keyed by a field, which often makes the subsequent processing much clearer.

rrebollo (approved):
Code Review. LGTM!

edescalona (approved):
LGTM

jelenapoblet:
@crrodrigueztrujillo we still need testing and a functional review to confirm this. From a technical perspective, I think we're good to go. @dkendall @johnotoole-binhex , could you please review it at your convenience?

johnotoole-binhex (approved):
Functional review ok

l10n_us_sales_tax_engine/models/us_tax_calculation_log.py
crrodrigueztrujillo: Thanks — you're right, `grouped()` expresses this much more clearly and keeps the same batching (one `browse()` per model). Since #173 is already merged I'll fold it into the next engine PR rather than reopening this one. The refactor is:

```python
@api.depends("res_model", "res_id")
def _compute_document_ref(self):
    self.document_ref = ""
    targets = self.filtered(lambda log: log.res_model and log.res_id)
    for res_model, logs in targets.grouped("res_model").items():
        docs = self.env[res_model].browse(logs.mapped("res_id")).exists()
        names = {doc.id: doc.display_name for doc in docs}
        for log in logs:
            log.document_ref = names.get(log.res_id) or f"{log.res_model},{log.res_id}"
```

Appreciated — noting `grouped()` for the rest of the codebase too.

PR #170, opened 2026-04-20
[19.0][MIG] l10n_us_base_county: Migration to 19.0

BT-pkosnik:

PR #174, opened 2026-05-24
[19.0][MIG] l10n_us_base_county: Migration to 19.0

dnplkndll:
Port of l10n_us_base_county to 19.0. Takeover of #170 with squashed fixes and linting resolved.


crrodrigueztrujillo (changes requested):
Thanks for the port! The CI is fully green. A few things need to be addressed before I can approve:

## Blocker

**`xlrd` not declared in `external_dependencies`** (`models/res_country_state_county.py:3`)

`xlrd` is imported at the top level of the model file, meaning Odoo cannot load the module at all if `xlrd` is not installed. Please add to `__manifest__.py`:

```python
"external_dependencies": {"python": ["xlrd"]},
```

## Should fix

1. **`get_state_id` called twice** in `_import_counties` — `state_id` is already assigned before the duplicate check; the `create()` call should reuse it instead of calling `get_state_id(state_code)` again.

2. **Wrong view record ID** — `view_county_form` contains a `<search>` arch, not `<form>`. Should be `view_county_search`.

3. **Duplicate `name` in `ir.model.access.csv`** — both access rules share `name = access_res_county_state_county`. The admin rule should use a distinct name, e.g. `access_res_country_state_county_system`.

4. **`state_id` field missing `ondelete`** on the County model — recommend `ondelete="restrict"`.

5. **`.pot` header says `Odoo Server 18.0`** — this is a 19.0 migration; the translation file header should reflect 19.0.

## Minor (no need to block on these)

- `maintainer` key in manifest is not a standard Odoo/OCA field; the info is already in `readme/CONTRIBUTORS.md`
- County model missing `_order` (defaults to `id` instead of alphabetical)
- `xlwt` test dependency not declared in any `requirements.txt`

PR #182, opened 2026-06-18
[18.0][ADD] l10n_us_attomdata: Paid US property data API.

amh-mw:
https://api.developer.attomdata.com/docs


amh-mw:
Formerly https://github.com/OCA/vertical-realestate/pull/21

jelenapoblet (commented):
@amh-mw  This looks interesting. Could you elaborate on the intended use case(s)? I'm trying to understand where this fits within l10n_us.

My understanding is that this API is primarily a property enrichment service for the US real estate market (ownership, valuations, deeds, permits, etc.). Is the goal to support a specific localization requirement, or is it intended as a reusable integration for industries such as real estate, property management, lending, or insurance?

amh-mw:
> @amh-mw This looks interesting. Could you elaborate on the intended use case(s)?

This is a paid service that one of my home improvement customers uses to enrich leads; determining whether or not a site will have construction restrictions due to historic preservation district, EPA lead safety due to property age, etcetera.

> I'm trying to understand where this fits within l10n_us.
> My understanding is that this API is primarily a property enrichment service for the US real estate market (ownership, valuations, deeds, permits, etc.).

I originally made this pull request against vertical-realestate...

> Is the goal to support a specific localization requirement, or is it intended as a reusable integration for industries such as real estate, property management, lending, or insurance?

This is all of the code that I felt was generic enough to be generally useful. I have a custom module with a couple hardcoded fields to bring the data in the last mile. I don't currently have the extra bandwidth to write any sort of broadly applicable API data to field mapping user interface.

PR #183, opened 2026-06-19
[18.0][IMP] l10n_us_sales_tax_engine: per-jurisdiction booking, FIPS, sourcing, single-local

dnplkndll:
Improvements to `l10n_us_sales_tax_engine` on top of #173, rebased onto `18.0` now that #173, #196, #197 and #199 have landed.

The engine collapsed every jurisdiction's rate into one combined `account.tax` per state. That computes the right total, but the GL then has no record of how much belongs to the state vs the county vs a transit district — which is exactly what a state return asks for. This books one child tax per non-zero component instead.

## Non-mechanical adaptations worth flagging

- **`_process` signature.** #199 added `partner`; this branch previously added its own `partner_id`. Dropped ours and resolve `partner.commercial_partner_id` inside `_process` instead — one parameter, and certificates keep resolving to the legal entity.
- **Exempt-source handling is now a union.** `exempt_partner` (#199), `exempt_customer` (the hook below) and `exempt_nexus` all take the whole-document path, on both the sale-order and invoice sides.
- **`account.tax.country_id` is NOT NULL in 18.0.** Booking sets it explicitly (company fiscal country → country → US). Worth noting that the engine's current tests only pass because CI loads demo data, which supplies a fiscal country; this makes them independent of that.
- **`us.tax.calculation.log.source` gains `single_local`.** This is a gap in this branch, not a pre-existing one — the single-local path is new here, and the first cut added the source without the matching Selection value, so the log write raised `ValueError` at confirm time for any electing seller. It survived because the tests exercised `_single_local_for` / `_apply_single_local_rate` in isolation and never reached the log write; there is now one that drives `_process` end to end.
- **Sourcing and single-local are state-configurable, not Texas-specific.** `add_on` covers the TX election; `combined` covers Alabama's SSUT flat rate, which replaces state + local rather than adding to it.

## Validation

Replayed 314 real transactions across 39 states against a commercial engine's output. Every line came back jurisdiction-tagged; combined-rate agreement is 64% using only free ZIP-level rate data, and the residual is ZIP-estimate-vs-rooftop divergence in the source data rather than engine logic — swapping in a rooftop provider for one state took it from 5/11 to 11/11 there.


dnplkndll:
@crrodrigueztrujillo I am posting this adaptation on your l10n_us_sales_tax_engine add to show the couple of adaptations. let me know how the discussion on the module _base adaptation goes as well.  I started to look at what that might look like. https://github.com/ledoent/l10n-usa/pull/19 

For me the shaping of the base module helps to have more of the use cases so I will PR here the rest of the concept to show why these adaptations where added. for smb there must surely be a complicity versus accuracy. but in most cases I would guess a good public repo like this should be able to resolve any state issue with a flexible public forum to post them. I was goign to try and tackle CA next but let me know if you have any state specific or exemption type to review. the dms around the exemptions management I wanted to run down the docuseal [do](https://github.com/ledoent/l10n-usa/pull/2) versus the odoo sign app and dms. just seemed to have a little better consumer grade usability on mobile. 

cc @jelenapoblet

jelenapoblet:
@dnplkndll I think it would be better to first focus on getting the *_base module agreed upon and merged, as that will define the overall architecture.

Once the base is in place, we can review the additional improvements and the rest of the modules on top of it. That will make it much easier to evaluate each contribution in the right context.
cc: @crrodrigueztrujillo

crrodrigueztrujillo:
Thanks for the deep dive — the per-jurisdiction booking, sourcing, and single-local logic are genuinely valuable and I'd like to bring them in. One structural note though: this snapshot reintroduces ziptax/api_ninjas as direct imports inside the engine's `_provider_service_classes()` — that's the exact pattern we just split out into standalone addons (#196/#197) per the architecture discussion with @jelenapoblet/@rrebollo. Once you rebase on the current #173 (post-split, post today's review fixes), the new capabilities here should drop in cleanly. Also flagging independently of the architecture question: `_process()`/`_apply_to_sale_order`/`_apply_to_invoice` lost their `self.sudo()` calls — that breaks background calculation for users without the `us_tax_*` groups, which was an intentional design point in the original.

crrodrigueztrujillo:
@dnplkndll Re-checked the current head (`a50a831`) and both of my earlier points are addressed — thanks for turning those around:

- `_provider_service_classes()` now only imports `provider_local`; no ziptax/api_ninjas imports left inside the engine.
- `self.sudo()` is back in `_process`, `_apply_to_sale_order` and `_apply_to_invoice`, so background calculation works again for users without the `us_tax_*` groups.

One of your two findings holds up; the other I'd push back on.

`account.tax.country_id` — you're right, and I verified it. In 18.0 it's `required=True, store=True, precompute=True`, computed as `company_id.account_fiscal_country_id or company_id.country_id`. Neither `_get_or_create_state_tax` nor `_get_or_create_exempt_tax` sets it, so tax creation silently depends on the company having a fiscal country configured — a company with no chart of accounts installed would hit the NOT NULL constraint on first calculation. In practice any US seller has a US chart, which is why it never surfaced, but relying on an implicit precondition rather than setting the value is a real gap on my side. Setting it explicitly is the right call and touches nothing else.

`us.tax.calculation.log.source` / `single_local` — this one isn't a pre-existing bug. `single_local`, `_single_local_for` and `_apply_single_local_rate` don't exist in `18.0`; the whole single-local sourcing path is new in this branch, so the missing Selection value is a gap in this PR's own coverage rather than something the merged tests overlooked. Worth rewording in the description — as written, a reviewer goes looking for it in the engine and doesn't find it.

Dropping your own `partner_id` parameter in favour of resolving `commercial_partner_id` inside `_process` is the right call either way; it's the same conclusion #200 reaches from the other direction.

On sequencing — @jelenapoblet, a question rather than a position. When you suggested focusing on the `_base` module first, #173 was still open. Looking at what actually merged: the engine no longer imports any provider (`_provider_service_classes()` only pulls in `provider_local`), and ziptax and api_ninjas ship as independent addons that declare `l10n_us_sales_tax_engine` in `depends`. Functionally that *is* the base + provider-plugins architecture you and @rrebollo asked for — the only thing it doesn't carry is a `_base` suffix in the module name.

So: does #173 as merged satisfy the base-module requirement? If it does, I'd rather unblock @dnplkndll now than have him hold eleven branches waiting on a refactor that's already done.

If you do want the literal rename, it's worth naming the cost first: `l10n_us_sales_tax_engine` is already released on PyPI, so a rename means a new package, a migration for existing installs, and updating `depends` in #196 and #197 — all for a naming convention, with no behavioural gain. My read is that it's not worth it in a stable series, but it's your call as maintainer and I'll follow it either way.

Answering your questions:

**California** — right next target. It has more multi-jurisdiction ZIPs than any other state, and your own 5/11 → 11/11 result with a rooftop provider is the evidence that ZIP-level data structurally can't get there. CDTFA is the correct source.

**Exemption types to prioritise** — resale, government, nonprofit and manufacturing cover the overwhelming majority of real B2B exemptions. That's essentially what your SST reason taxonomy in #186 already models, so I think you're on the right list already.

**DocuSeal vs Odoo Sign** — my strong preference is that neither goes in the tax modules. Storing the certificate as an `ir.attachment` behind a hook keeps the fiscal logic independent of the signature stack, and anyone who wants e-signature layers it on top. Binding a tax module to DocuSeal (external, non-OCA) or to Odoo Sign (Enterprise) locks out a large share of users for something orthogonal to tax calculation. OCA has `sign_oca` for those who want it in-tree. Keep the hook, skip the opinion.

Once the sequencing question above is settled I'll review the rebase properly — the per-jurisdiction booking and single-local work is the part I most want to see land.

jelenapoblet:
My original concern was ensuring a clean separation between the tax engine and the provider implementations.

IMO naming is important for keeping things well structured, but at this point the architecture is more important, and that separation appears to be in place.

Unless there's another technical reason to split or rename the module, I wouldn't block the remaining PRs on that point alone.

ericlakich:
Disclosure: I work at TaxCloud, which runs Ziptax (zip.tax). One of the providers in this repo is ours, so weigh that accordingly. @crrodrigueztrujillo invited input on this PR over in #198, so here it is. Most of it is design commentary, and I have kept the part that is actually about us at the end so it is easy to skip.

## The gap I would most want closed before this merges: rate versioning

Per-jurisdiction booking is the right call and the GL argument for it is correct. A state return asks for the split, so the split has to exist in the ledger. But creating one `account.tax` per component surfaces a problem the combined-rate approach hid: local and district rates change, and they change on a cadence. Many states adjust local rates on quarterly boundaries, so a component that is 0.5% in one quarter can be 0.75% in the next.

If the per-component equivalent of `_get_or_create_state_tax` resolves an existing `account.tax` by jurisdiction and then treats that record's `amount` as the authority, the first calculation after a rate change either:

- edits the existing record, silently restating the rate behind every already-posted invoice pointing at it, or
- creates a near-duplicate with a colliding name, which is a reporting mess of a different kind.

Two ways out, and I think the second fits this codebase better:

1. **Effective-dated tax records.** Put the rate version in the identity so `KS Allen County 1.0%` and `KS Allen County 1.25%` are distinct records, and look up by jurisdiction plus document date. Correct, but `account.tax` grows without bound.
2. **Snapshot the resolved component rate on the line at post time** and treat `account.tax` as a label for the jurisdiction rather than the authority on the rate. `us.tax.calculation.log` is already most of the way there. The payoff is that recalculating an old document reproduces the old number instead of today's number, which is what someone asking about a 2024 invoice two years from now actually needs.

Either way the thing to avoid is a mutable rate on a long-lived record. I could not tell from the description which side the current head lands on, so apologies if this is already handled.

## On the FIPS question in ledoent/l10n-usa#24: yes, but not as the jurisdiction identity

Your argument for an unprefixed `fips_code` on the county is right, and `avatax_code` is the correct comparison to draw. FIPS is a federal standard rather than a vendor identifier, and namespacing it under a sales-tax module would force the next consumer to depend on a tax module for a geography fact.

Where I would push back is on how far it gets carried:

- **Your own Kansas number is the ceiling.** Special districts have no FIPS, so 66% of jurisdictions there cannot be keyed this way. If FIPS becomes the join key for *booking*, districts need a parallel scheme immediately and you have two identity systems in one table.
- **The identifier a return wants is usually the state's own locality code**, not FIPS. States publish their own local codes and the return form is keyed to those. FIPS is excellent for joining to Census and other federal datasets, which is the actual case you make in #24, but I would not expect to file against it.

So: add `fips_code` as a geography attribute on the county, use it for the naming fix and the demographic join, and keep jurisdiction identity as its own opaque key that can also hold a state-assigned code for districts. That banks the 34% naming win without committing the other 66% to a scheme that cannot represent them.

One caveat on the naming fix itself: whether a given state's published SST files carry usable county names alongside `fips_county` varies by state and by file. Worth checking the specific state's file rather than assuming the join is always required, since in some cases the name is present and the `KS County 001` placeholder is a loader artifact rather than missing source data.

## On the exemption metadata ask from #198, I think the answer is that we should not carry it

@crrodrigueztrujillo asked whether the ziptax provider should communicate exemption metadata in the API request. My honest answer is no, and the current design is already right.

Exemption is a decision about an entity and a certificate, not about a location. Resolving it in the engine at Step 3.5, before any provider call, keeps the exemption record and its audit trail in Odoo where the certificate actually lives, and it makes the behaviour identical regardless of which provider is configured. Pushing it into the provider request would make exemption handling vary by vendor, which is the opposite of what a provider abstraction is for.

The only thing worth passing to a provider is whatever it needs to resolve the jurisdiction, so an exempt sale still lands on the right state and locality for the exempt-sales line on the return. It sounds like `exempt_partner` already carries `state_id`, which covers that.

## What I can offer, if useful

California being next is where I think we can help without anyone taking our word for anything: a fixture set of real addresses in multi-jurisdiction ZIPs with the resolved component breakdown for each, so the per-jurisdiction booking tests run against real splits rather than synthetic ones. Happy to produce it as a plain data file with no dependency on us, usable by `provider_local` or any other provider. If that is wanted I would open it separately rather than adding scope here.

Caveat so the offer is fair: California district taxes are not a fixed depth. Some are city-level, some county-wide with city carve-outs, and some overlap, so a component list for CA is not the same shape as Kansas. Worth verifying per district rather than assuming a four-slot state/county/city/special model holds everywhere. CDTFA is the right source, as already noted above.

## Honest limits on us

Ziptax is rate lookup. No filing, no returns, and product taxability mapping is on the caller, so it does not replace what the engine here is doing. For most of the design questions in this PR you do not need a paid provider at all. The place one matters is the accuracy gap already measured upthread, which is one narrow slice of this work.

Ignore whichever parts of this do not fit. You both know this codebase and OCA conventions far better than I do.

jelenapoblet:
I think there are several good suggestions here that we should consider, as long as they remain provider-agnostic.
Regarding account.tax, I agree there's an architectural problem to solve, but I don't think creating a new tax record every time a rate changes is the right solution. @crrodrigueztrujillo @dnplkndll do you have any ideas on how we could address this?

rrebollo:
@jelenapoblet, you asked for my review, but this PR is still in draft state. So, are we really ready to start the review process, @dnplkndll?

crrodrigueztrujillo:
@jelenapoblet @dnplkndll sorry for the delay. I went through this against core before answering, because the whole question hinges on one assumption: that changing `account.tax.amount` restates posted invoices. In 18.0 it doesn't. `_sync_tax_lines` skips anything that is not draft, `account.move.line.write()` raises if `tax_ids`/`tax_line_id` change on a posted line, and `_compute_totals` depends on the `tax_ids` link, not on `amount`. Posted tax lines are frozen amounts; `account.tax` is already a label in core's own model. So I agree with Jelena: we should not create a tax record per rate version.

What I'd propose, provider-agnostic:

1. **`account.tax` = one stable record per jurisdiction**, identity = `(company, us_tax_level, us_tax_state_id, us_tax_jurisdiction_id)` — the tag fields this PR already adds — and looked up by those fields, not by name. Name without the percentage (`US Sales Tax FL Alpha County`). Only the anonymous level-only components (what API providers return, no jurisdiction id) keep the rate in their identity, otherwise `US Sales Tax TX County` would flip between Harris and Travis on every calculation.
2. **`amount` = the rate currently in force.** When `us.tax.rate(doc_date)` differs from `tax.amount`, the engine writes the new amount. Posted documents are untouched (see above); the group taxes are already keyed by composition so they become stable too — zero growth.
3. **The authority on rate history is `us.tax.rate`** (already effective-dated) plus a per-line snapshot of `rate_detail` in `us.tax.calculation.log` — today the log only stores the document-level best rate, so a per-line JSON is what lets us reproduce a 2024 invoice in 2027.
4. **Back-dated documents** (the one real gap): when `rate(doc_date) != tax.amount` the engine writes the historical amount on the tax line before posting — manual tax lines are first-class in core and `_sync_tax_lines` keeps them — or, as a minimum, logs `source="rate_mismatch"`.

For this PR that means: `_get_or_create_component_tax` looks up by the tag fields and updates `amount` when it differs; `_get_or_create_group_tax` stays as is; `_log` stores the per-line detail; and a test for a rate change on an existing jurisdiction (there is none in `test_jurisdiction_booking.py`). `test_sale_order.py` asserting `"US Sales Tax NY 0%"` would need adjusting.

Two other things I found while reviewing that I'd consider blocking, separately from the above:

- `learn_jurisdiction` stores learned rooftops with `zip` and `confidence=1.0`, and `get_best_jurisdiction`'s ZIP fallback doesn't filter `address_key = ""` — so one learned rooftop in a ZIP that crosses jurisdictions becomes the ZIP-level answer for every other address in that ZIP (`order="confidence desc"`). That inverts the problem the feature is meant to fix.
- Origin-based intrastate sourcing: `_sourcing_zip` sends the origin ZIP but the payload still carries the destination `address`/`city`, and `resolve_jurisdiction` tries the destination `address_key` first — so origin sourcing is overridden as soon as the destination has a learned rooftop. `test_sourcing.py` doesn't cover rooftop + origin.

Smaller: `_get_or_create_group_tax` scans every group tax of the company in Python per line; the tag "heal" writes `us_tax_jurisdiction_id: False` when triggered by level only; `_get_or_create_exempt_tax` still lacks `country_id` (from the earlier thread). Happy to re-review once it's out of draft.

@ericlakich on your other three points: agreed on FIPS — geography attribute on the county, not the jurisdiction identity (the identity stays an opaque key, which is also what the proposal above relies on); agreed on exemption staying in the engine at Step 3.5 rather than traveling in provider requests — that's how #199 landed and it stays that way; and yes to the California fixture offer — real multi-jurisdiction addresses as a plain data file, in its own PR, would be genuinely useful for testing the booking split. Please open it when ready.

PR #189, opened 2026-06-19
[18.0][ADD] l10n_us_sales_tax_provider_taxjar: TaxJar provider as a standalone addon

dnplkndll:
TaxJar rate provider in its own addon, registering via the engine's documented `_provider_service_classes` extension point - the `engine_base + provider_xxx` direction (ziptax/api_ninjas extract identically).

Depends on #183 (the engine).

> Draft for upstream review; `[DO NOT MERGE]` pin resolves the in-flight engine PR (#183).


jelenapoblet:
@crrodrigueztrujillo

crrodrigueztrujillo:
@jelenapoblet thanks for the ping — reviewed against `18.0` and against the two merged provider addons. The skeleton follows the ziptax pattern (`_provider_service_classes` via `super()`, whool, readme fragments, tests on `UsTaxBaseTest`), so structurally it's the right shape. Points, most important first:

1. **Hard code dependency on #183**: `COUNTY_NAME_KEY`/`CITY_NAME_KEY` and `self._named_jurisdictions(...)` (`services/provider_taxjar.py`) only exist in #183's `provider_base.py`. On current `18.0` the addon installs, but every `get_rate` raises `AttributeError` — which is not a `ProviderError`, so the engine doesn't even fall through to the next provider. Either this waits for #183 or those three lines (and `test_taxjar_uses_lowercase_name_keys`) go, which is trivial — the rate part is independent.
2. `test-requirements.txt` pinned to `refs/pull/183/head` must be removed before merge.
3. `data/provider_taxjar.xml`: add `noupdate="1"` (as `us_tax_provider_ziptax.xml`); otherwise every `-u` resets `active=False` / `sandbox_mode=True` on a live provider. And default `sandbox_mode` to `False` like ziptax/api_ninjas — the sandbox host needs its own token, so a production token 401s out of the box.
4. No `res_config_settings` toggle / `config_parameter="l10n_us_tax.taxjar_token"` (password) / `_sync_provider_active("taxjar", …)` inside `us_tax_providers_block`, mirroring `l10n_us_sales_tax_provider_ziptax/models/res_config_settings.py`. Without it the token has no UI.
5. `normalize_response`: TaxJar returns `combined_district_rate` — use it instead of deriving district by subtraction with a clamp to 0, and keep the 1:1 mapping the other two providers use.
6. `get_rate`: reject an empty ZIP with `ProviderError` (ziptax does), and since `SUPPORTS_ADDRESS=True` send `street`/`country=US` too — or set it to `False` and `supports_address_level=False` in the seed. `requests.ConnectionError` isn't caught either (same gap as ziptax, worth closing here).
7. Manifest `category` → `"Accounting/Localizations"`; and please double-check the author/copyright header ("Binhex - Carlos R. Rodriguez" on `provider_taxjar.py`) — fine by me if that's the intent since the code originated in the engine, just make sure it's deliberate. An engine-level fallback test with `@mute_logger` like `test_provider_ziptax.py` would round it out.

Once #183 is settled this should be a quick one.

PR #196, opened 2026-06-26
[18.0][ADD] l10n_us_sales_tax_provider_ziptax: ZipTax provider for l10n_us_sales_tax_engine

crrodrigueztrujillo:
## Summary

Follow-up to #173 (the base engine). Adds the **ZipTax (zip.tax)** provider
addon, as agreed with @jelenapoblet and @rrebollo on the base PR.

- Self-registers with `l10n_us_sales_tax_engine`'s provider registry via
  `_provider_service_classes()` — no edit to the engine, installs/uninstalls
  independently of any other provider
- ZIP-level rate lookups, free tier of 100 calls/month
- Settings toggle + API key under *Accounting → Settings → US Tax API
  Providers*, with a "Test Credentials" action

## Dependency

Depends on:
- [ ] https://github.com/OCA/l10n-usa/pull/173

A temporary reference to #173 was added in `test-requirements.txt`
(commit "[DON'T MERGE] test-requirements.txt") so `test with Odoo`/
`test with OCB` can resolve `l10n_us_sales_tax_engine` before #173 merges,
per the OCA convention: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

This intentionally fails `Detect unreleased dependencies` as a do-not-merge
signal — that commit will be dropped once #173 merges.

## Test plan
- [x] Install on top of `l10n_us_sales_tax_engine`, no errors
- [x] Settings shows the ZipTax API key field, "Test Credentials" validates
      against the real zip.tax API
- [x] Fallback confirmed against the real API (not mocked): a ZIP with no
      local rate data falls back to ZipTax and returns a real combined rate
- [x] Uninstall — base module stays intact


crrodrigueztrujillo:
Note on CI: \`test with Odoo\` / \`test with OCB\` will stay red until #173 (the base engine) is merged — this branch is based on \`18.0\` and doesn't contain \`l10n_us_sales_tax_engine\` yet, so \`oca_install_addons\` can't resolve the dependency. That's expected, not a bug in this PR; \`pre-commit\` and \`Detect unreleased dependencies\` are the checks that matter here before #173 merges.

edescalona (approved):
LGTM

jelenapoblet (commented):
Overall, I think the architecture looks good. The provider is nicely decoupled from the engine and follows the extension pattern we designed, so I don't see anything that would block the merge.

I only have two small suggestions:

ZIPTAX_BASE_URL and ZIPTAX_SANDBOX_URL currently point to the exact same endpoint. Unless ZipTax provides a dedicated sandbox URL in the future, I would remove the unused constant to avoid confusion.
If we eventually add more providers (TaxJar, Avalara, Vertex, API Ninjas, etc.), it could be worth centralizing the HTTP request logic in ProviderBase (timeouts, retries, session handling, error mapping). It's not necessary for this PR, but it would reduce duplicated code as the provider ecosystem grows.

Other than that, the implementation looks clean to me and I don't have any architectural concerns. I'd be happy to merge it.

crrodrigueztrujillo:
Thanks for the thorough review, @jelenapoblet!

**ZIPTAX_SANDBOX_URL**: removed — you're right, the constant was defined but never referenced (both call sites already used `ZIPTAX_BASE_URL`). Pushed in the latest amend.

**HTTP centralization in ProviderBase**: fully agree on the long-term value. Since #173 is already merged, touching `ProviderBase` now would require a separate follow-up PR against the base. I'd prefer to keep that improvement scoped to a dedicated issue once all three providers are in, so it can be designed consistently across ziptax, api_ninjas, and any future providers.

jelenapoblet (approved):

johnotoole-binhex (approved):
Functional review ok

PR #197, opened 2026-06-26
[ADD] l10n_us_sales_tax_provider_api_ninjas: API Ninjas provider for l10n_us_sales_tax_engine

crrodrigueztrujillo:
## Summary

Follow-up to #173 (the base engine), in the same series as #196 (ZipTax).
Adds the **API Ninjas (api-ninjas.com)** Sales Tax provider, as agreed with
@jelenapoblet and @rrebollo on the base PR.

- Self-registers with `l10n_us_sales_tax_engine`'s provider registry via
  `_provider_service_classes()` — no edit to the engine, installs/uninstalls
  independently of any other provider
- ZIP-level rate lookups, free tier available — county/city/special rates
  return the literal string "This field is for premium subscribers only" on
  the free tier; `normalize_response()` falls back to summing the available
  numeric components instead of erroring out
- Settings toggle + API key under *Accounting → Settings → US Tax API
  Providers*, with a "Test Credentials" action

## Dependency

Depends on:
- [ ] https://github.com/OCA/l10n-usa/pull/173

A temporary reference to #173 was added in `test-requirements.txt`
(commit "[DON'T MERGE] test-requirements.txt") so `test with Odoo`/
`test with OCB` can resolve `l10n_us_sales_tax_engine` before #173 merges,
per the OCA convention: https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference(s)-to-another-pull-request(s)

This intentionally fails `Detect unreleased dependencies` as a do-not-merge
signal — that commit will be dropped once #173 merges.

## Test plan
- [x] Install on top of `l10n_us_sales_tax_engine`, no errors
- [x] Settings shows the API Ninjas key field, "Test Credentials" validates
      against the real api-ninjas.com API
- [x] Fallback confirmed against the real API (not mocked): San Francisco CA
      (no local coverage, ZipTax disabled to force this path) returned
      `state_rate: 6%`, `source: api` — county/city correctly 0 (free tier
      limitation)
- [x] Installed alongside ziptax — priority respected, no conflicts
- [x] Uninstall — base module stays intact


edescalona (approved):
LGTM

l10n_us_sales_tax_provider_api_ninjas/tests/test_provider_api_ninjas.py
rrebollo, line 1: Have you considered mocking the entire test suite? Wouldn't this exhaust the free tier limits because of GitHub's OCA CI/CD?

johnotoole-binhex (approved):
Functional review ok

jelenapoblet (approved):
Looks good to me, let's merge it

l10n_us_sales_tax_provider_api_ninjas/tests/test_provider_api_ninjas.py
crrodrigueztrujillo: Good concern, and fully agreed on the principle — CI must never consume a provider quota. The suite here is already fully mocked, so the OCA CI never reaches api-ninjas.com:

- `test_get_rate_parses_real_response_shape` patches `requests.get` with a `FakeResponse` built from a real free-tier payload (including the literal `"This field is for premium subscribers only"` strings), so it's a pure parsing/regression test.
- `test_local_to_api_ninjas_fallback` patches both `ProviderLocal.get_rate` and `ProviderApiNinjas.get_rate` — the engine's fallback chain is exercised without any HTTP.
- `test_api_ninjas_registered_in_provider_service_classes` only touches the registry.

On top of that, no API key is configured in CI, so even an accidental live call would fail fast rather than burn quota. `l10n_us_sales_tax_provider_ziptax` follows the same pattern.

The confusion probably comes from my Test plan wording ("confirmed against the real API (not mocked)") — that referred to the **manual QA** I ran locally with a real key, not to the automated tests. Sorry for the ambiguity; I'll phrase that as "manual QA (live API)" in future provider PRs. Resolving this thread unless you see something I missed.

Issue #198, opened 2026-07-27
l10n_us_sales_tax_engine: partner-level tax exemption and shipping address selection


crrodrigueztrujillo:
## Summary

Following the merge of #173 (`l10n_us_sales_tax_engine`), two closely related
features are missing that are standard in US sales tax compliance workflows:

1. **Partner-level tax exemption** — mark a customer as tax-exempt with an
   exemption code and certificate number (resale, government, nonprofit, etc.)
2. **Shipping address toggle** — choose whether tax jurisdiction is determined
   by the billing address or the delivery address on a per-order basis
   (US tax law generally taxes at the point of delivery)

Both are present in Avalara/AvaTax's Odoo integration as reference implementation.

## Current behavior

- The engine has no mechanism to exempt a specific customer from taxes.
  Exemption today is only possible via: (a) no nexus in the customer's state,
  or (b) product category set to `EXEMPT`.
- The engine always resolves the tax address from the sale order's invoice
  address (`partner_invoice_id`). There is no toggle to use the shipping
  address (`partner_shipping_id`) instead.

### 1. `res.partner` — new fields (new `models/res_partner.py`)

- `us_tax_exempt` — Boolean, "Is Tax Exempt"
- `us_tax_exemption_code` — Selection (RESALE, GOVERNMENT, NONPROFIT, AGRICULTURAL, INDUSTRIAL, EDUCATIONAL, OTHER)
- `us_tax_exemption_number` — Char, certificate number issued by the state

Displayed in a new **Tax Exemption** section inside the existing `Sales & Purchase` tab (no new tab needed for a few fields).

### 2. `sale.order` — new fields (extend existing `models/sale_order.py`)

- `us_tax_based_on_shipping` — Boolean, default `True`, "Tax based on shipping address"
- `us_tax_exemption_number` — Char, computed from partner but overridable per order
- `us_tax_exemption_code` — Selection, same, overridable per order

### 3. `services/tax_engine.py` — new Step 0

Before the nexus check, the engine would short-circuit if the partner is exempt:

```python
# Step 0: Partner-level exemption — no provider call needed
if partner and partner.us_tax_exempt:
    return {"source": "exempt_partner", "tax_amount": 0.0, "lines": []}
```

Address resolution would also respect the new toggle:

```python
if order.us_tax_based_on_shipping and order.partner_shipping_id:
    address = resolve_address(order.partner_shipping_id)
else:
    address = resolve_address(order.partner_invoice_id)
```

[...]

PR #199, opened 2026-07-27
[IMP] l10n_us_sales_tax_engine: partner tax exemption + shipping address toggle

crrodrigueztrujillo:
## Summary

Implements issue #198 — two complementary features requested in the original PR review and modeled after AvaTax's UX:

### 1. Partner-level tax exemption certificate
- New fields on `res.partner`: **Is Tax Exempt** (Boolean), **Exemption Code** (e.g. `RESALE`, `GOVERNMENT`), **Exemption Certificate #**
- New section injected in the *Sales & Purchase* tab of the partner form
- Engine inserts a new step (Step 3.5) **after nexus check**: if the partner holds a certificate, every line is exempted with the explicit 0 % "US Sales Tax - Exempt" tax and the audit log records `source = exempt_partner`
- Logic: no nexus → `exempt_nexus` (unchanged); nexus exists + partner exempt → `exempt_partner`; otherwise → standard per-line calculation

### 2. Tax Based on Shipping Address toggle
- New Boolean field `us_tax_based_on_shipping` on `sale.order` (default `True` — preserves existing behavior)
- When `False`, the engine resolves the **billing/invoice address** instead of the shipping address for the tax calculation
- New helper `resolve_invoice_address()` extracted from `address_resolver.py` (shared logic via private `_partner_to_address()`)

### Other changes
- `us_tax_calculation_log.source` Selection: added `exempt_partner` value
- Version bumped to `18.0.1.1.0`
- 5 new tests in `test_partner_exemption.py`

## Test plan
- [ ] `test_exempt_partner_short_circuits_engine` — exempt partner → `source = exempt_partner`, `tax_amount = 0`
- [ ] `test_non_exempt_partner_in_fl_calculates_tax` — standard customer in nexus state → tax calculated normally
- [ ] `test_shipping_address_toggle_uses_invoice_address` — toggle OFF + NY billing → `exempt_nexus` (no nexus in NY)
- [ ] `test_shipping_address_toggle_default_uses_shipping` — toggle ON (default) → FL shipping → tax calculated
- [ ] `test_exempt_partner_audit_log_created` — audit log entry with `source=exempt_partner` is written

cc @jelenapoblet @rrebollo — this is the feature discussed in issue #198


jelenapoblet:
@crrodrigueztrujillo is this a new module or a fix to an existing one ? would it work also with other Sale Tax related addons like Avatax ?

crrodrigueztrujillo:
Hi @jelenapoblet! This is a feature improvement to the existing `l10n_us_sales_tax_engine` module merged in #173 — not a new module.

Regarding AvaTax compatibility: the two features added here are provider-agnostic by design:

1. **Partner exemption** — the `us_tax_exempt` check on `res.partner` happens at **Step 3.5** in the engine orchestrator, *before* any provider call. If a partner is exempt, the engine short-circuits and returns `exempt_partner` without invoking any provider. This means AvaTax (or any other provider following the registry pattern) would automatically benefit from it.

2. **Shipping address toggle** — `us_tax_based_on_shipping` on `sale.order` controls address resolution *before* the address is passed to the provider. Any provider receives the already-resolved address, so the toggle is equally transparent to AvaTax.

The only touch point with AvaTax would be at the `res.partner` fields level — AvaTax uses the same concept (exemption certificate on the customer) so the fields are aligned with its data model. If the AvaTax provider (tracked in #183) is eventually merged, it would consume `us_tax_exempt` / `us_tax_exemption_number` from the same fields.

rrebollo (commented):
First of all, I think the technical implementation proposed in this PR is sound. You clearly have strong expertise in both the tax computation services being integrated and the underlying business domain, so I don't have concerns from a technical standpoint.

My input is mainly from a design perspective. I don't mean to suggest the current approach is incorrect, but rather to raise a few questions that came up during the review. I don't necessarily have the answers, but I hope they help validate or refine the design.

Regarding the questions in the linked issue:

### Question 1

Based on the described scope, I believe this functionality fits well within the current add-on. Keeping it here seems consistent with the module's responsibility.

### Question 3 (Invoices)

From my experience, this functionality naturally extends beyond Sales and should also apply to Accounting/Invoicing.

Whether this should remain in the same add-on or be split is a separate discussion, but I would strongly recommend checking existing OCA patterns before deciding.

For instance, discount-related modules (like triple or global discount) come to mind. If I recall correctly, they are implemented primarily on accounting models and then extended into Sales via mixins or inheritance.

Also, Sales already depends on Accounting, so both flows are tightly connected by design. That relationship is important when evaluating where this logic should live.

I would suggest aligning with existing OCA implementations in Accounting and Sales rather than introducing a new pattern.

### Question 2 (Exception Code Values)

This is where my design concerns are most relevant.

You clearly have deep knowledge of US tax rules and the external engines involved, which is essential for this integration.

However, regardless of how those external systems model their concepts, the key question is how to translate them into Odoo’s functional model rather than reproducing them as-is.

For example—and I may be wrong here, since this is not my area of expertise—in Europe tax exemptions are typically handled via Fiscal Positions rather than directly marking a partner as tax-exempt. The partner is assigned a Fiscal Position, and that determines tax behavior.

This is why I have some doubts about introducing tax exemption data directly on the partner. Odoo usually models this indirectly through Fiscal Positions.

My intuition is that a similar principle might apply in the US localization as well, but I’m not certain, so I’m framing this as a question rather than a recommendation.

More generally, I think this is the key design principle for this integration: when external tax engines introduce concepts that don’t exist in Odoo, we should first check whether they can be expressed using existing Odoo mechanisms before adding new ones.

The goal is to preserve Odoo’s architecture and conventions while still supporting the required functionality, translating external concepts into Odoo’s language rather than importing them directly.

Again, I may be mistaken here, but I wanted to share the observation as it stood out during the review.

### Shipping Address Toggle (Taxation Based on Delivery Address)

The second feature in the PR—the shipping address toggle for taxation—should likely be evaluated under the same reasoning.

I have even less domain knowledge here, so I may be missing constraints, but I suspect this also requires some level of Odoo-side adaptation.

Still, I would apply the same principle: first check whether existing Odoo concepts already cover this behavior (or something close), and only introduce a new mechanism if strictly necessary.

So I wouldn’t reject the idea at all—it likely makes sense—but I would frame it within the same design approach: prefer existing Odoo models first, extend only when needed.

---

Overall, my impression is that the implementation is technically correct. My comments are mainly about design consistency and alignment with existing Odoo and OCA patterns, and I hope they are helpful in evaluating the final approach.

crrodrigueztrujillo:
Hi @rrebollo, thank you for the thoughtful and well-structured review — this is exactly the kind of design discussion that makes OCA contributions stronger.

**On Fiscal Positions (Question 2)**

You're absolutely right that Fiscal Positions are the canonical Odoo mechanism for tax exemptions, and I want to be clear: I considered this path before going with direct partner fields. The reason I didn't use Fiscal Position as the primary mechanism is specific to how external tax APIs work.

Providers like AvaTax, TaxJar, and the ones already integrated in this module require the exemption **certificate number** to be sent as part of the API request itself — not just a 0%-tax result applied after the fact. If we relied solely on a Fiscal Position to zero out taxes on the Odoo side, the external API would still return a positive amount, and we'd be silently overriding it. This breaks the audit trail and doesn't satisfy US tax authority documentation requirements, which expect the exemption reason to be recorded at the point of calculation.

So the fields on `res.partner` (`us_tax_exemption_code`, `us_tax_exemption_number`) serve as metadata carriers for the API call, not as a replacement for Fiscal Position logic. The short-circuit at Step 3.5 is a performance and correctness optimization — avoiding an unnecessary API roundtrip when we already know the result is zero.

That said, your observation opens a valid evolution path: using a dedicated "US Tax Exempt" Fiscal Position as the *trigger* while keeping the certificate fields on the partner for API metadata. This hybrid approach would be more aligned with Odoo's architecture. I'm happy to explore it if the reviewers consider it a blocker — or we could track it as a follow-up improvement.

**On Invoices (Question 3)**

This PR already covers the Accounting flow. `calculate_for_invoice` on `us.tax.engine.service` was updated in this same PR to pass `partner=move.partner_id`, so the Step 3.5 exemption check applies equally to `account.move` records. A dedicated test (`test_exempt_partner_on_invoice`) covers this path.

**On the Shipping Address Toggle**

Odoo already has `partner_shipping_id` vs `partner_invoice_id` on `sale.order`, but the standard tax computation always uses the customer's Fiscal Position (billing-side), regardless of where the goods are delivered. The `us_tax_based_on_shipping` toggle exposes a choice that genuinely doesn't exist in Odoo standard — it's required because US Sales Tax is a **destination-based** tax in most states: the rate is determined by where the goods are *delivered*, not where the customer is billed. There's no existing Odoo mechanism that covers this, so a new field is necessary.

Happy to discuss any of these points further. Thanks again for the review.

rrebollo:
In that case, I'd suggest implementing a hybrid approach.

crrodrigueztrujillo:
Hi @rrebollo, fully agree — the hybrid approach is the right long-term design direction. Let me share our thinking on where that work belongs.

**The fields in this PR are the foundation, not the final pattern**

The three fields added to `res.partner` (`us_tax_exempt`, `us_tax_exemption_code`, `us_tax_exemption_number`) are the minimal data layer the engine needs to function. They don't replace Fiscal Positions — they carry API metadata that Fiscal Positions cannot express on their own (the certificate number must be sent as part of the provider API request, as I explained in my previous comment). These fields are the foundation that any further integration — including a Fiscal Position trigger — would build on top of.

**The hybrid approach belongs in PR #183**

Looking at the existing work, PR #183 (`l10n_us_sales_tax_exemption` by @dnplkndll) is already scoped exactly for this: building the exemption management layer on top of the engine. It already introduces a `_get_customer_exemption()` hook designed for extensibility. Adding a `is_us_tax_exempt` flag to `account.fiscal.position` and wiring it to Step 3.5 is precisely the kind of work that belongs there — not in the engine base module.

The engine's Step 3.5 is already designed to be extended: a future PR can add a second check that reads the fiscal position without touching anything in this PR. Both trigger mechanisms (direct partner flag + fiscal position) can coexist cleanly.

**Why this PR should not be blocked on that**

In the US, **over 40–50% of B2B transactions involve some form of tax exemption** (resale, government, manufacturing, nonprofit). Without this PR, the engine cannot be used in production by any company selling to other businesses. This is not an edge case — it is the dominant real-world use case.

Blocking the merge because the trigger convention (direct flag vs. Fiscal Position) isn't the most Odoo-native option means prioritizing form over function on something that is critical for daily business operations. Rolando, you yourself noted the implementation is technically correct — the only open question is which convention we prefer for the trigger.

**Our proposal**

- Merge PR #199 as-is: the foundation is correct and covers the core business case.
- Track the Fiscal Position integration in PR #183, which is already scoped for the exemption management layer.

If you consider the hybrid approach a hard blocker for #199, please let us know exactly what to implement here and we will do it. Otherwise, we'd like to move forward and handle the Odoo-native integration pattern in #183.

Thanks again for the constructive discussion — this is exactly the kind of review that makes OCA modules better.

rrebollo (approved):
Code Review. LGTM!

jelenapoblet (approved):
Functional review ok

PR #200, opened 2026-07-29
[18.0][FIX] l10n_us_sales_tax_engine: delegate partner exemption to the commercial entity

dnplkndll:
An exemption certificate is issued to the legal entity, not to each of its contacts. `#199` reads `us_tax_exempt` from the order's partner directly, so a sale shipped to a child contact of an exempt customer is taxed in full — the certificate lives on the parent only.

## Why `_commercial_fields()` rather than a check in the engine

`vat` and `company_registry` already delegate this way: they belong to the commercial entity and Odoo propagates them to its contacts. An exemption certificate number is the same class of field, so adding the three fields to `_commercial_fields()` fixes the resolution at the source — Step 3.5 is untouched, and any future consumer of `us_tax_exempt` (an order, an invoice, a provider payload) gets the corrected value for free.

The alternative — resolving `commercial_partner_id` inside the engine — would fix only the call site that remembered to do it.

## Reproducing

Create a contact under a partner with **Is Tax Exempt** set, then confirm an order for that contact: before this change the order is fully taxed and the audit log records a rated source instead of `exempt_partner`.

Covered by two tests: a delivery contact created under an exempt customer is exempt and its order books no tax, and flagging an existing customer exempt propagates to contacts already on file.

### DIFF
--- a/l10n_us_sales_tax_engine/README.rst
+++ b/l10n_us_sales_tax_engine/README.rst
@@ -151,6 +151,7 @@ Contributors
 ------------
 
 - Carlos R. Rodriguez <c.rodriguez@binhex.cloud>
+- Don Kendall dkendall@ledoweb.com
 
 Maintainers
 -----------
--- a/l10n_us_sales_tax_engine/__manifest__.py
+++ b/l10n_us_sales_tax_engine/__manifest__.py
@@ -2,7 +2,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 {
     "name": "US Sales Tax Engine",
-    "version": "18.0.1.2.0",
+    "version": "18.0.1.2.1",
     "category": "Accounting/Localizations",
     "summary": (
         "Hybrid USA Sales Tax engine: local DB first, "
--- a/l10n_us_sales_tax_engine/migrations/18.0.1.2.1/post-migrate.py
+++ b/l10n_us_sales_tax_engine/migrations/18.0.1.2.1/post-migrate.py
@@ -1,0 +1,33 @@
+# Copyright 2026 Ledo Enterprises
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+EXEMPTION_FIELDS = [
+    "us_tax_exempt",
+    "us_tax_exemption_code",
+    "us_tax_exemption_number",
+]
+
+
+def migrate(cr, version):
+    """Sync exemptions set before the fields were delegated.
+
+    A post_init_hook would not reach these: Odoo calls it only for a new
+    install, and on a new install no partner carries the field yet.
+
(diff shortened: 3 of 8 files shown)


crrodrigueztrujillo:
Thanks for catching this — the bug is real and the diagnosis is right. Reading `us_tax_exempt` off the document's partner meant a sale shipped to a child contact of an exempt customer was taxed in full, and `_commercial_fields()` is the correct place to fix it: it's the same mechanism Odoo uses for `vat` and `company_registry`, and an exemption certificate belongs to the same class of data — issued to the legal entity, valid for its contacts. Fixing it at the source rather than resolving `commercial_partner_id` inside the engine is the better call, for exactly the reason you give: every future consumer of the field inherits the correction without having to remember to do it.

Two things I'd like to see addressed before this goes in:

**1. Records already on file.** `_commercial_fields` only syncs on `create` and on `write` to the parent. Contacts that already exist under a partner flagged exempt stay unsynced until someone happens to write on that parent — and since #199 is already merged and released, there are installations that may have the flag set from before this fix. `test_exemption_added_later_reaches_existing_contacts` covers the write path, not the upgrade path. A `post_init_hook` that syncs existing data would close it:

```python
def post_init_hook(env):
    exempt = env["res.partner"].search([("us_tax_exempt", "=", True)])
    exempt._commercial_sync_to_children()
```

**2. The fields stay editable on child contacts.** `views/res_partner_views.xml` renders the *US Tax Exemption* group on every partner form, with no condition on `parent_id`. After this change a user can edit the certificate on a child contact and the next write on the parent silently overwrites it. Worth making them readonly when the partner isn't its own commercial entity, so the form doesn't offer an edit that won't stick — something like `readonly="commercial_partner_id != id"` on the three fields (or hiding the group outright on child contacts).

Neither is a reason to rework the approach — (1) is a hook and (2) is a view attribute. Happy to approve once they're in, or to push them myself as a follow-up if you'd rather keep this PR minimal. Thanks for the tests, they're exactly the two cases that matter.

dnplkndll:
agreed on the migration and read only for child addresses when not equal to commercial id.

crrodrigueztrujillo (changes requested):
Thanks — the `_commercial_fields()` delegation, the `readonly="commercial_partner_id != id"` on the form and the two new tests look right to me, and CI is green.

One blocker in `migrations/18.0.1.2.1/post-migrate.py`: `exempt._commercial_sync_to_children(fields_to_sync=EXEMPTION_FIELDS)` is called on the whole recordset. Inside core, `_commercial_sync_to_children` does `commercial_partner = self.commercial_partner_id` and then `_update_fields_values()` reads `self[fname]` on it — with two or more exempt customers that is a multi-record set and the upgrade fails with `ValueError: Expected singleton`. The tests don't exercise the migration, so CI can't catch it.

Suggested fix:
```python
exempt = env["res.partner"].search(
    [("us_tax_exempt", "=", True), ("parent_id", "=", False)]
)
for partner in exempt:
    partner._commercial_sync_to_children(fields_to_sync=EXEMPTION_FIELDS)
```
(restricting to commercial entities also avoids re-running the sync from children that were already copied). With that in, I'm happy to approve.

PR #202, opened 2026-08-13
[18.0][IMP] l10n_us_sales_tax_engine

edescalona:
@BinhexTeam

T24775

New setting to define whether US Taxes calculation is run automatically.

### The problem

Tax is only calculated when someone presses **Calculate US Tax**, or on confirm/post. Change a quantity, a price or the customer address afterwards and the tax is stale, with nothing on screen saying so: the quotation the customer sees can carry a tax that does not apply. And the automation that exists cannot be separated from the engine — `l10n_us_tax.engine_active` is the only control.

### The setting

**Automatic Tax Calculation** (`us_tax_auto_calculate` → `l10n_us_tax.auto_calculate`, default False), in Settings → Accounting → US Sales Tax Engine.

Ticked, the engine also runs on its own: on `write()`/`create()` of a draft document or its lines, on order confirmation, on invoice posting, and on two daily crons. Unticked, no automatic trigger fires and only the manual button calculates — narrower than `18.0.1.2.0`, where confirm and post recalculated as long as the engine was on. Unticking the engine turns the setting off with it.

What cannot be recalculated in place is flagged instead: change the ZIP of a customer with 300 draft quotations and none is recalculated inside your transaction — each shows a banner saying the tax is outdated, with a button to recalculate.

### On upgrade

The setting is **inherited from the engine switch**, so an update never silently stops calculating tax: engine already on → automation on. A database that never enabled the engine is untouched. Be aware the automation goes further than `18.0.1.2.0` (write/create + daily cron), so it consumes more provider quota — untick it for the previous behaviour. `post-migrate.py` stamps the fingerprint on already-calculated documents so outdated detection covers them too. Neither script spends a provider call.

### How outdated tax is detected

A SHA-256 fingerprint of exactly what the engine reads (ZIP, state, city, country, exemption, date, company, and each line's fiscal category and subtotal), compared against the fingerprint of the last attempt. That comparison is the only trigger guard, so no field allowlist exists: writing `note`, writing the same value twice, or rewriting `33101` as `33101-4521` reaches nothing. Outdated documents are fixed by the next edit, the banner button, or the daily crons in batches of 200.

### Scope

`_us_tax_state_domain()` — one domain serving both the write guard and the cron's `search()`:

- `sale.order`: `state in (draft, sent, sale)`, not locked, `invoice_status not in (invoiced, upselling)`.
- `account.move`: draft customer invoices and refunds.

Outside that the fingerprint is frozen, banner and buttons are hidden, and nothing touches the document.

### Behaviour changes and fixes

- **An address with no country is no longer taken for an American one.** A partner with a ZIP and a state but empty **Country** used to be taxed by the button; the engine now answers `skip_non_us` and applies nothing. Set the country on those partners.
- The **Block** failure policy applies to the manual button only. On the automatic path the error is logged and swallowed, so a provider outage cannot make every draft document impossible to edit, confirm or post.
- A failed calculation leaves the document uncalculated: the tax is rolled back to a savepoint, the audit row records the error, and the document stays outdated for the cron. A document whose *first* calculation failed is outdated like any other.
- **Calculate US Tax** is hidden when the engine is disabled, and stamps the document as calculated only when a calculation actually happened — not when the engine is off, not when the engine skips the document as non-US.
- Documents outside `_us_tax_state_domain()` are never rewritten by a customer address change: posted, cancelled, locked and fully invoiced documents, and vendor bills of the same partner.
- A calculation runs once per change. `us_tax_skip_auto` in the context makes the engine's own writes on the lines return at the trigger's first guard, so applying the tax cannot re-enter the automation before the calculated hash is stamped.
- One cron record per document model. `ir.cron._callback` rolls back the whole job on failure, so a `sale.order` error in a shared record would discard the `account.move` work of the same run. A downstream addon adds a third document type by shipping its own record.


edescalona:
Hi @jelenapoblet @rrebollo @johnotoole-binhex, when you can, please do a review. Any suggestions are welcome.

l10n_us_sales_tax_engine/models/account_move.py
```diff
@@ -18,16 +22,131 @@ class AccountMove(models.Model):
         string="Tax Calculated At",
     )
 
+    us_tax_input_hash = fields.Char(
+        string="US Tax Input Hash",
+        compute="_compute_us_tax_input_hash",
+        store=True,
+        readonly=True,
```
rrebollo, line 29:
```suggestion
```

rrebollo: computed-stored fields are readonly by default

l10n_us_sales_tax_engine/models/account_move.py
```diff
@@ -18,16 +22,131 @@ class AccountMove(models.Model):
...
+        "partner_id.country_id",
+        "partner_id.us_tax_exempt",
+        "partner_shipping_id",
+        "partner_shipping_id.zip",
+        "partner_shipping_id.state_id",
+        "partner_shipping_id.city",
+        "partner_shipping_id.street",
+        "partner_shipping_id.country_id",
+        "invoice_line_ids.price_subtotal",
+        "invoice_line_ids.product_id.us_tax_category_id",
```
rrebollo, line 63: Given the large number of fields triggering the recomputation, if we keep this approach, I would use a callable for `@api.depends`, since it would make the implementation more dynamic and easier to maintain. However, I think the number of dependencies also points to a broader design issue. If a third-party addon further down the dependency hierarchy changes something that should affect this value, it seems counterintuitive that we would need to know about it and modify our code to explicitly add another trigger. This introduces coupling that we probably want to avoid.

I wonder whether OCA's [component_event](https://github.com/OCA/connector/tree/18.0/component_event) could be worth exploring. An event-driven approach might decouple recomputation from explicit field and addon dependencies. This feels more like a design issue than a technical one, and there may be a simpler, more robust mechanism worth considering before sticking with the current approach; if we do, I would at least switch `@api.depends` to a callable.

l10n_us_sales_tax_engine/models/account_move.py
```diff
@@ -48,8 +167,7 @@ def _post(self, soft=True):
         posted, its lines' tax_ids can no longer be modified ("you should
         reset the journal entry to draft to do so").
         """
-        ICP = self.env["ir.config_parameter"].sudo()
-        if ICP.get_param("l10n_us_tax.engine_active", "False") == "True":
+        if self._us_tax_auto_enabled():
             for move in self.filtered(
                 lambda m: m.move_type in ("out_invoice", "out_refund")
```
rrebollo, line 172:
```suggestion
                lambda m: not m.is_sale_document())
```

l10n_us_sales_tax_engine/models/account_move_line.py
```diff
@@ -0,0 +1,37 @@
...
+    def _us_tax_auto_moves(self):
+        """Return the parent moves of the product lines in self.
+
+        account.move.line.create() and write() run inside the core's
+        _sync_dynamic_lines/_sync_tax_lines, which creates, writes and
+        deletes tax and counterpart lines. Only the lines the engine
+        actually prices — the ones in move_id.invoice_line_ids — may
+        trigger a recalculation.
+        """
+        return self.filtered(lambda line: line in line.move_id.invoice_line_ids).move_id
```
rrebollo, line 19: I don't understand. In the end, you're returning an `account.move`, so why not just do:

```suggestion
        return self.mapped('move_id')
```

l10n_us_sales_tax_engine/models/account_move_line.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Recalculate the parent moves that just got new invoice lines."""
+        lines = super().create(vals_list)
+        lines._us_tax_auto_moves()._us_tax_auto_recalculate()
```
rrebollo, line 25: I don't get it. You're also triggering recomputation in the `create` and `write` methods of `account.move`. How are you avoiding recursion?

From a design perspective, I find this approach a bit risky - might be the right word. That said, I'm aware things have gotten messy in the accounting domain. Still, this only strengthens the case for adopting a component‑based event approach instead.

l10n_us_sales_tax_engine/models/sale_order.py
```diff
@@ -26,26 +34,137 @@ class SaleOrder(models.Model):
...
+        string="US Tax Calculated Hash",
+        readonly=True,
+        copy=False,
+        help="Value of the input hash at the time of the last calculation.",
+    )
+    us_tax_is_stale = fields.Boolean(
+        string="US Tax Outdated",
+        compute="_compute_us_tax_is_stale",
+        help="The tax inputs changed after the last US tax calculation.",
+    )
```
rrebollo, line 55: If you already have some mixins in place, why not include these fields there—and the same goes for account.move as well?

l10n_us_sales_tax_engine/models/sale_order.py
```diff
@@ -26,26 +34,137 @@ class SaleOrder(models.Model):
...
 
     def action_confirm(self):
-        """Auto-calculate tax on confirmation if engine is active."""
-        ICP = self.env["ir.config_parameter"].sudo()
-        if ICP.get_param("l10n_us_tax.engine_active", "False") == "True":
+        """Auto-calculate tax on confirmation if the automation is enabled."""
+        if self._us_tax_auto_enabled():
             for order in self:
                 try:
                     self.env["us.tax.engine.service"].calculate_for_sale_order(order)
```
rrebollo, line 170: These are pretty similar to the ones in `account.move`. You might find some common ground to include them in a mixin, perhaps with some dynamic input parameters.

l10n_us_sales_tax_engine/models/sale_order_line.py
rrebollo, line 1: mixin???

l10n_us_sales_tax_engine/models/us_tax_auto_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        The batch limit caps how many documents each run recalculates;
+        the rest are picked up by the following run.
+        """
+        if not self._us_tax_auto_enabled():
+            return
+        for model_name in ("sale.order", "account.move"):
+            model = self.env[model_name]
+            domain = model._us_tax_state_domain() + [
+                ("us_tax_calculated_hash", "!=", False)
+            ]
```
rrebollo, line 118: That seems odd. Polymorphism should take care of this, so why are you using `self.env[model_name]`? Since those models already inherit this mixin, what's the point? Shouldn't you not have to worry about that? Could you clarify?

rrebollo (commented):
Thank you for your contribution. Any thoughts on my suggestions?

edescalona:
Hi @rrebollo, thank you for your comments, I'll review them and get back to you.

crrodrigueztrujillo:
Maintainer view, after going through the branch and Rafael's comments:

The stored fingerprint + comparison is a good direction and I'd keep it — I don't think `component_event` buys us anything here (it also fires from `write()`, and pulling OCA/connector into a localization is a heavy dependency). But the current shape has a few functional problems I'd want fixed before merge:

- `action_confirm`/`_post` call the engine directly, without `us_tax_skip_auto` and without stamping `us_tax_calculated_hash`, so the line-level `tax_id` writes re-trigger `_us_tax_auto_recalculate` (and on SO the `date_order=now()` write changes the hash again) — two or three engine runs per confirm, each one a log entry and possibly an API call. The tests with the setting ON don't count calls.
- `invoice_date` is part of the hash but core sets it in `_post` after we stamped — every posted invoice ends up `us_tax_is_stale=True`, with a banner whose "Recalculate" can't work on a posted move.
- `@api.depends("partner_id.zip", …)` recomputes the hash of every document of the partner, posted or not, inside the partner's transaction; and the `sale` state in the domain lets the cron rewrite taxes on invoiced/locked orders.
- The exception swallow has no savepoint, and the hash is stamped even when the engine fails.
- Behaviour change on upgrade: existing installs with the engine active stop calculating on confirm/post until the new setting is enabled — needs a CHANGELOG note or a pre-migrate copying the flag.

On Rafael's points I agree with all of the structural ones: one mixin holding fields/computes/`write`/`action_calculate_us_tax`/cron with the concrete models only implementing `_us_tax_get_address/_doc_date/_lines/_engine_run/_state_domain`, `@api.depends` with a callable so third-party addons can extend the fingerprint, and one `ir.cron` per model instead of `self.env[model_name]`. `us_tax_is_stale` stored and the cron searching with `limit` instead of `filtered` in Python. Then tests asserting exactly one calculation on confirm and on post with the setting ON, and that posting doesn't leave the invoice stale.

@edescalona could you pick up @rrebollo's inline comments from Aug 14 together with the above? They've been open for a while and most of them point at the same refactor, so one round should cover both.

edescalona:
ping @rrebollo @crrodrigueztrujillo

jelenapoblet:
Addressing Rafael’s comments together with these points sounds like the right way forward 👍

